PR #5, opened 2024-11-11
catch exceptions in max score compute

xingdi-eric-yuan:
For cases where `re.search(r"collected (\d+) items", output)` is None.

### DIFF
--- a/froggy/utils.py
+++ b/froggy/utils.py
@@ -262,10 +262,7 @@ def cleanup_pytest_output(output):
 
 def extract_max_score_from_pytest_output(output):
     # ... collected 25 items
-    try:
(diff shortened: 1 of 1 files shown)


froggy/utils.py
```diff
@@ -262,7 +262,10 @@ def cleanup_pytest_output(output):
 
 def extract_max_score_from_pytest_output(output):
     # ... collected 25 items
-    return int(re.search(r"collected (\d+) items", output).group(1))
+    try:
+        return int(re.search(r"collected (\d+) items", output).group(1))
+    except:
+        return 1
```
MarcCote:
```suggestion
    return int(re.search(r"collected (\d+) items?", output).group(1))
```

PR #7, opened 2024-11-13
Fix Context Truncation

xingdi-eric-yuan:


froggy/envs/swe_bench.py
```diff
@@ -35,7 +35,7 @@ def instructions(self):
 
     def load_dataset(self):
         self.ds = load_hf_dataset(self.HF_SWE_BENCH_VERIFIED)["test"]
-        self.dataset = sorted(self.ds["instance_id"])
+        self.dataset = {item: "" for item in sorted(self.ds["instance_id"])}
```
MarcCote: What's this?

xingdi-eric-yuan: env.dataset should be a dict.

MarcCote: Ok, but what is the value supposed to be if self.dataset is supposed to be a dict?

PR #10, opened 2024-11-14
toolbox

mormio:


froggy/tools/toolbox.py
```diff
@@ -0,0 +1,45 @@
...
+from pathlib import Path
+from typing import Type, Dict, Optional, Any, Callable
+
+class Toolbox:
+    _tool_registry: Dict[str, Type] = {}
+
+    @classmethod
+    def register(cls, name: str=None, config_cls: Optional[Any] = None) -> Callable:
+        def decorator(subclass: Type) -> Type:
+            nonlocal name
```
MarcCote: Is that from MTTL codebase? Do you actually needs nonlocal here?

froggy/tools/toolbox.py
```diff
@@ -0,0 +1,45 @@
...
+            return subclass
+
+        return decorator
+
+    @classmethod
+    def load_tools(cls):
+        tools_dir = Path(__file__).parent
+        for tool_file in tools_dir.glob("*.py"):
+            if tool_file.stem not in ["__init__", "registry"]:
+                importlib.import_module(f"froggy.tools.{tool_file.stem}")
```
MarcCote: Not sure if we actually need this. As long as the import structure is sound, it will register all the tools as they are discover by the imports. (I think)

froggy/tools/toolbox.py
mormio: It's not from the MTTL codebase, I used nonlocal for name because it allows CodePatcher to be referred to as patcher I think...

MarcCote: Got it. I see it now. Otherwise, it would throw errors?

froggy/tools/toolbox.py
MarcCote: See [ddb64a7](https://github.com/microsoft/Froggy/pull/10/commits/ddb64a7e97b3f4f689c7dbf6232eebba892bb17d)

froggy/tools/toolbox.py
MarcCote: Actually, if you use a different name for the local variable, that will work. I'm on the fence with the `nonlocal` keyword. E.g.
```python
    @classmethod
    def register(cls, name: str=None, config_cls: Optional[Any] = None) -> Callable:
        def decorator(subclass: Type) -> Type:
            name_ = name or subclass.__name__.lower().replace('tool', '')
            if name_ in cls._tool_registry:
                if subclass != cls._tool_registry[name_][0]:
                    raise ValueError(
                        f"Cannot register '{name_}' multiple times."
                    )
                return subclass

            cls._tool_registry[name_] = (subclass, config_cls)
            subclass.registered_name = name_
            return subclass

        return decorator
```

froggy/tools/toolbox.py
mormio: I think you're right! Pushed

PR #13, opened 2024-11-19
Terminal refac

matheper:
This WIP PR contains:
- Decouple the terminal operations from environments and tools.
- New `Terminal` class handles terminal operations, including running commands and managing pseudo terminals (interactive sessions).
- New `DockerTerminal` to handle operations within Docker containers. `Terminal` and `DockerTerminal` should work interchangeably.
- Refactoring of `pdb.py` to use `Terminal` class
- Adding tests for `Terminal`, `DockerTerminal`, and PDBTool classes

TODO:
- [x] Run docker with non-root user
- [x] Provide the option to initialize the terminal using docker or the regular terminal (done through config)
- [x] Clean up docker containers (implemented using atexit, will revisit that in the future)
- [ ] Update remaining tools and envs that use a terminal (I'll do that in a separate pr)

### DIFF
--- a/froggy/envs/swe_bench.py
+++ b/froggy/envs/swe_bench.py
@@ -37,8 +37,7 @@ class SWEBenchEnv(RepoEnv):
         self.dataset = {row["instance_id"]: row for row in self.ds.sort("instance_id")}
 
     def reset(self, *, seed=None, options: dict = None):
-        if options is None:
-            options = {}
+        options = options or {}
         assert "task_name" in options, "task_name must be provided in options"
         assert (
             options["task_name"] in self.dataset
--- a/froggy/envs/terminal_simulator.py
+++ b/froggy/envs/terminal_simulator.py
@@ -28,8 +28,7 @@ class TerminalSimulatorEnv(RepoEnv):
         self.tempdir_name = self.tempdir.name
 
     def reset(self, *, seed=None, options: dict = None):
-        if options is None:
-            options = {}
+        options = options or {}
         self.tempdir.cleanup()
         self.current_sample = self.dataset[options["task_name"]]
         shutil.copytree(self.bug_free_code_path, self.tempdir_name, dirs_exist_ok=True)


froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+class Terminal:
+
+    def __init__(
+        self,
+        working_dir: str = None,
+        setup_commands: list[str] = None,
+        env_vars: dict[str, str] = None,
+    ):
+        if working_dir is None:
+            working_dir = "/tmp/Froggy"
```
MarcCote: Should we use `tempfile.TemporaryDirectory(prefix="RepoEnv-Terminal-")`?

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+            os.makedirs(working_dir, exist_ok=True)
+        self.setup_commands = setup_commands if setup_commands else []
```
MarcCote:
```suggestion
        self.setup_commands = setup_commands or []
```

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+        self.env_vars = env_vars if env_vars else {}
```
MarcCote:
```suggestion
        self.env_vars = env_vars or {}
```

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+
+        command = [entrypoint] if isinstance(entrypoint, str) else entrypoint
+
+        if self.setup_commands:
+            command = self.setup_commands + ["&&"] + entrypoint
+
+        logger.debug(f"Running command in terminal: {command}")
+        process = subprocess.Popen(
+            command,
+            env=self.env_vars | self.path_env,
```
MarcCote: Should we also keep the os.environ ones?
```suggestion
            env=self.env_vars | self.path_env | dict(os.environ),
```

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+        # Clean up output by disabling terminal prompt and colors
+        self.env_vars["NO_COLOR"] = "1"  # disable colors
+        self.env_vars["PS1"] = ""  # disable prompt
+        self.working_dir = working_dir
+        self._master = None  # PTY master file descriptor
+
+    @property
+    def path_env(self):
+        # TODO: find a better way to set PATH
+        return {"PATH": f"{os.path.dirname(sys.executable)}:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}
```
MarcCote: Why is this needed?

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def run(self, entrypoint, working_dir=None, timeout=None):
```
MarcCote: When would we want to change the working_dir instead of using the one defined when creating the Terminal instance?

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+        )
+        return success, output
+
+    def run_interactive(
+        self, entrypoint: str, expected_output: str = "", timeout: int = 30
+    ):
+        """Run a command in the interactive terminal and return the output.
+        Requires a PTY. The terminal stays open after the command is executed.
+        """
+        # TODO: include working_dir parameter to align with run()?
```
MarcCote: See comment for run(). In any case, I agree we should match the signature of run().

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+            stdout=slave,
+            stderr=slave,
+            text=True,
+            close_fds=True,
+            start_new_session=True,
+        )
+
+        # close slave, end in the parent process
+        os.close(slave)
+        atexit.register(self.close_pseudo_terminal)
```
MarcCote, line 147: Good call.

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+        If no expected_output is provided, read for timeout seconds.
+        """
+        output = ""
+        start_time = time.time()
+        while True:
+            if not expected_output and time.time() - start_time > timeout:
+                logger.debug("Timeout reached while reading from PTY.")
+                break
+            try:
+                # [TODO] 1024? We might need more than that.
```
MarcCote:
```suggestion
```

froggy/terminal.py
matheper: I think never. I'm ok removing this parameter.

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+                - ``mode`` Either ``rw`` to mount the volume read/write, or
+                  ``ro`` to mount it read-only.
+        """
+        super().__init__(
+            working_dir=working_dir,
+            setup_commands=setup_commands,
+            env_vars=env_vars,
+        )
+        self.base_image = base_image
+        self.volumes = volumes if volumes else {}
```
MarcCote:
```suggestion
        self.volumes = volumes or {}
```

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+            command,
+            workdir=working_dir or self.working_dir,
+            environment=self.env_vars,
+            stdout=True,
+            stderr=True,
+        )
+        success = status == 0
+        return success, output.decode().strip("\r\n").strip("\n")
+
+    def clone(self) -> Terminal:
```
MarcCote:
```suggestion
    def clone(self) -> "DockerTerminal":
```

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+        terminal = self.__class__(
```
MarcCote: Any reason to not make it explicit?
```suggestion
        terminal = DockerTerminal(
```

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+
+    @property
+    def default_entrypoint(self) -> list[str]:
+        """Starts a new bash session exporting the current python executable as 'python'.
+        Flags --noprofile and --norc are used to avoid loading any bash profile or rc file,
+        which could interfere with the terminal setup (clean outputs)"""
+        return ["/bin/bash", "--noprofile", "--norc"]
+
+    def clone(self) -> "Terminal":
+        return self.__class__(
```
MarcCote: Any reason to not make it explicit?
```suggestion
        return Terminal(
```

froggy/terminal.py
matheper: The current entrypoints are "python -m pdb...", "pytest ...". When we execute those commands with Popen, they use the system python, not python from the virtual/conda environment.

matheper: I don't like this solution. If you have any suggestion, I'm happy to try it.

froggy/terminal.py
```diff
@@ -0,0 +1,308 @@
...
+            setup_commands=self.setup_commands,
+            volumes=self.volumes,
+            env_vars=self.env_vars,
+            working_dir=self.working_dir,
+        )
+        return terminal
+
+    def setup_container(self):
+        # Create and start a container mounting volumes and setting environment variables
+        suffix = "".join(random.choices(string.ascii_lowercase, k=8))
```
MarcCote: What happens it random has been seeded? It would cause to container to have the same name. Do you think there's a way of letting docker generate the unique name and maybe rename it to add "froggy-..." as prefix?

froggy/terminal.py
matheper: That's a good question. I think the class should not do that by default.
We can either add a flag to load them automatically, or pass them when instantiating the class.

froggy/tools/eval.py
```diff
@@ -13,7 +13,7 @@ class EvalTool(EnvironmentTool):
     }
 
     def register(self, environment):
-        from froggy.envs import RepoEnv
+        from froggy.envs.env import RepoEnv
```
MarcCote, line 16: good catch. that might be caught in another PR though.

froggy/terminal.py
matheper: Doesn't work very well to be honest... 😂
A context manager (with statement) would probably work better.

froggy/tools/eval.py
MarcCote: Nope, just checked. all Good.

froggy/tools/pdb.py
```diff
@@ -61,24 +54,65 @@ def __init__(
         self.persistent_breakpoints = persistent_breakpoints
         self.auto_list = auto_list
         self.current_frame_file = None
+        self._terminal: Terminal = None
```
MarcCote, line 57: is it really optional?

froggy/terminal.py
MarcCote: Why? For instance, if you activate a conda env, doesn't it change the ENV Vars (e.g., pointing to the right Python path)?

froggy/terminal.py
MarcCote: I think that's related to not using os.environ

froggy/terminal.py
MarcCote: Great

froggy/terminal.py
MarcCote: I agree with statement are more robust. I just couldn't find a good flow to leverage it for the working dir, so that's why I was using this atexit. Anyhow, it's probably better than nothing.

froggy/terminal.py
```diff
@@ -38,16 +41,25 @@ def __init__(
...
 
+    def prepare_command(self, entrypoint: list[str]) -> list[str]:
+        """Prepares a shell command by combining setup commands and entrypoint commands.
+        Then wraps the command in a shell (self.default_entrypoint) call."""
         if self.setup_commands:
-            command = self.setup_commands + ["&&"] + entrypoint
+            entrypoint = " && ".join(self.setup_commands + entrypoint)
+        else:
+            entrypoint = " && ".join(entrypoint)
+        command = shlex.split(
```
MarcCote, line 61: Didn't know about shlex. That's a standard lib?

froggy/terminal.py
```diff
@@ -274,35 +299,71 @@ def clone(self) -> Terminal:
...
+        image_tag = f"{base_image}-{self.host_uid}-{self.host_gid}"
+        try:
+            self.docker_client.images.get(image_tag)
+            logger.debug(f"Image {image_tag} already exists.")
+        except docker.errors.ImageNotFound:
+            logger.debug(f"Building image {image_tag}.")
+            dockerfile_bytes = dockerfile.encode("utf-8")
+            context = io.BytesIO()
+            with tarfile.open(fileobj=context, mode="w") as tar:
+                tarinfo = tarfile.TarInfo("Dockerfile")
```
MarcCote: It needs to be a tar-type of file?

froggy/terminal.py
matheper: Makes sense. I added the flag `include_os_env_vars` to control when to include the os env vars. By default, it's true, but one can decide to disable it, for example in `DockerTerminal`.

froggy/terminal.py
matheper: Right, path is correct when using os env vars. Removed.

froggy/terminal.py
matheper: That's the exact same reason I'm using atexit. I will keep it for now.

froggy/terminal.py
matheper: Yes, it is: https://docs.python.org/3/contents.html

froggy/terminal.py
matheper: Good catch, it doesn't. It's using BytesIO now.

froggy/tools/eval.py
matheper: Both work since `RepoEnv` is imported in `froggy/envs/__init__.py`. In this case, importing from `froggy.envs` was causing a circular dependency error.

froggy/terminal.py
matheper: At some point `DockerTerminal` was not implementing this method directly, but now it is. I'm changing that.

froggy/tools/pdb.py
matheper: Not really, it's required. However, the pdb terminal is only assigned at the environment reset.

froggy/terminal.py
matheper: Good point, thanks! Fixed.

froggy/envs/aider.py
```diff
@@ -30,7 +30,9 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.load_dataset()
 
-    def reset(self, *, seed=None, options={}):
+    def reset(self, *, seed=None, options: dict = None):
+        if options is None:
+            options = {}
```
MarcCote:
```suggestion
        options = options or {}
```

froggy/envs/env.py
```diff
@@ -62,30 +62,38 @@ def __init__(
...
+        readonly_patterns: list[str] = None,
         run_on_rewrite: bool = True,
         run_timeout: Optional[int] = None,
         dir_tree_depth: Optional[int] = None,
         auto_view_change: bool = True,
         terminal: Optional[Terminal] = None,
     ):
         """ """
+        if readonly_patterns is None:
+            readonly_patterns = []
```
MarcCote:
```suggestion
```

MarcCote: Actually. do we need it here?

froggy/envs/env.py
```diff
@@ -62,30 +62,38 @@ def __init__(
...
     def setup_workspace(
         self,
         path: str,
-        entrypoint: Optional[str] = None,
-        readonly_patterns: list[str] = [],
+        entrypoint: str,
+        readonly_patterns: list[str] = None,
     ):
+        if readonly_patterns is None:
+            readonly_patterns = []
```
MarcCote:
```suggestion
        readonly_patterns = readonly_patterns or []
```

froggy/envs/env.py
```diff
@@ -161,7 +169,9 @@ def restore(self, *filepaths):
 
             shutil.copy2(self.path / filepath, self.working_dir / filepath)
 
-    def reset(self, *, seed=None, options={}):
+    def reset(self, *, seed=None, options: dict = None):
+        if options is None:
+            options = {}
```
MarcCote:
```suggestion
        options = options or {}
```

MarcCote: I think it was fine as it was before, no? The code doesn't try to modify options.

froggy/envs/env.py
```diff
@@ -200,7 +210,7 @@ def reset(self, *, seed=None, options={}):
 
     def run(self):
         success, output = self.terminal.run(
-            self.entrypoint, self.working_dir, timeout=self.run_timeout
+            [self.entrypoint], timeout=self.run_timeout
```
MarcCote, line 209: Why the brackets?

froggy/envs/swe_bench.py
```diff
@@ -36,7 +36,9 @@ def load_dataset(self):
         self.ds = load_hf_dataset(self.HF_SWE_BENCH_VERIFIED)["test"]
         self.dataset = {row["instance_id"]: row for row in self.ds.sort("instance_id")}
 
-    def reset(self, *, seed=None, options={}):
+    def reset(self, *, seed=None, options: dict = None):
+        if options is None:
+            options = {}
```
MarcCote:
```suggestion
        options = options or {}
```

froggy/envs/env.py
```diff
@@ -61,20 +62,24 @@ def __init__(
         self,
         path: Optional[str] = None,
```
matheper:
```suggestion
        path: str | None = None,
```

froggy/envs/env.py
```diff
@@ -61,20 +62,24 @@ def __init__(
...
         entrypoint: str = "python -m pytest -sv .",
-        readonly_patterns: list[str] = [],
+        readonly_patterns: list[str] = None,
```
matheper:
```suggestion
        readonly_patterns: list[str] | None = None,
```

froggy/envs/terminal_simulator.py
```diff
@@ -27,7 +27,9 @@ def __init__(self, bug_free_code_path, buggy_code_path, **kwargs):
         self.tempdir = tempfile.TemporaryDirectory(prefix="TerminalSimulatorEnv-")
         self.tempdir_name = self.tempdir.name
 
-    def reset(self, *, seed=None, options={}):
+    def reset(self, *, seed=None, options: dict = None):
+        if options is None:
+            options = {}
```
MarcCote:
```suggestion
        options = options or {}
```

froggy/envs/env.py
```diff
@@ -61,20 +62,24 @@ def __init__(
...
         run_on_rewrite: bool = True,
         run_timeout: Optional[int] = None,
         dir_tree_depth: Optional[int] = None,
```
matheper:
```suggestion
        run_timeout: int | None = None,
        dir_tree_depth: int | None = None,
```

froggy/envs/env.py
```diff
@@ -61,20 +62,24 @@ def __init__(
...
         auto_view_change: bool = True,
+        terminal: Optional[Terminal] = None,
```
matheper:
```suggestion
        terminal: Terminal | None = None,
```

froggy/envs/env.py
matheper: terminal run requires a list of strings 😳

MarcCote: Nevermind, I parsed the line incorrectly in my head.

froggy/tools/pdb.py
MarcCote: Good point.

PR #14, opened 2024-11-19
Reset after Rewrite

xingdi-eric-yuan:
add option that resets prompt history after each rewrite (including the latest rewrite in history)

Also moving build_history_prompt and build_system_prompt (repeating) from each individual agent to base agent.

### DIFF
--- a/froggy/agents/utils.py
+++ b/froggy/agents/utils.py
@@ -72,38 +72,6 @@ class HistoryTracker:
         return len(self.memory)
 
 
-def load_config():
-    parser = argparse.ArgumentParser()
-    parser.add_argument("config_file", help="path to config file")
-    parser.add_argument("--agent", help="zero_shot, cot, tadpole", default="zero_shot")
-    parser.add_argument(
-        "--debug", action="store_true", help="Before sending action to the environment."
-    )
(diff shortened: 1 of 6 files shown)


froggy/agents/tadpole.py
```diff
@@ -21,71 +21,6 @@ def __init__(self, config_dict, env, verbose=False, _uuid=None):
         self.history = HistoryTracker(self.config["memory_size"])
```
MarcCote: Should the self.history go in the AgentBase then?

froggy/agents/utils.py
```diff
@@ -0,0 +1,247 @@
...
+    # Parse overriden params.
+    for param in args.params:
+        fqn_key, value = param.split("=")
+        entry_to_change = config
+        keys = fqn_key.split(".")
+        for k in keys[:-1]:
+            entry_to_change = entry_to_change[k]
+        entry_to_change[keys[-1]] = yaml.safe_load(value)
+    # print(config)
+    return config, args
```
MarcCote: I don't think that should be in agent.utils

MarcCote: I changed it back.

PR #15, opened 2024-11-21
zero shot agent that can access pdb after certain number of rewrite a…

mormio:
…ctions

### DIFF
--- a/froggy/agents/zero_shot.py
+++ b/froggy/agents/zero_shot.py
@@ -84,7 +84,6 @@ class AgentZeroShot_NoPDB(AgentZeroShot):
 
     def __init__(self, config_dict, env, verbose=False, _uuid=None):
         super().__init__(config_dict, env, verbose, _uuid)
-        self.history = HistoryTracker(self.config["memory_size"])
 
     def build_system_prompt(self, info):
         system_prompt = {}
@@ -122,7 +121,6 @@ class AgentZeroShot_PdbAfterRewrites(AgentZeroShot):
 
     def __init__(self, config_dict, env, verbose=False, _uuid=None):
         super().__init__(config_dict, env, verbose, _uuid)
-        self.history = HistoryTracker(self.config["memory_size"])
 
     def run(self, task_name=None, debug=False):
 
@@ -189,4 +187,5 @@ class AgentZeroShot_PdbAfterRewrites(AgentZeroShot):
                     )
                 )
                 break
+
         return done


froggy/agents/zero_shot.py
```diff
@@ -181,3 +181,73 @@ def build_question_prompt(self):
...
+
+            prompt = self.build_prompt(info)
+            answer, token_usage = self.llm(
+                prompt, info, temperature=self.config["llm_temperature"][0]
+            )
+
+            if debug:
+                breakpoint()
+
+            if info["rewrite_counter"] >= self.config["n_rewrites_before_pdb"] - 1:
```
mormio: this series of if are the main change

froggy/agents/zero_shot.py
```diff
@@ -181,3 +181,73 @@ def build_question_prompt(self):
...
+
+    def __init__(self, config_dict, env, verbose=False, _uuid=None):
+        super().__init__(config_dict, env, verbose, _uuid)
+        self.history = HistoryTracker(self.config["memory_size"])
+
+    def run(self, task_name=None, debug=False):
+
+        # remove the pdb tool from the environment
+        assert "pdb" in self.env.tools, "pdb not found in env tools"
+        pdb_tool = self.env.tools.pop("pdb")
```
mormio, line 129: start with pdb tool in the environment, then remove it to reinsert later.

MarcCote: clever.

froggy/agents/zero_shot.py
MarcCote: Thanks for pointing this out. Why the `- 1`? If `n_rewrites_before_pdb==1`, then you want at least one rewrite to have been done right? I.e. `info["rewrite_counter"] >= 1` not `info["rewrite_counter"] >=0`.

MarcCote: Oh! I see. You are adding it later if the current action is a rewrite. I'm assuming this is because the environment is the one listing the available tools for the next step.

MarcCote: The issue I see, is the rewrite might fail (i.e., wrongly formatted). :/

MarcCote: Ok, what about we add the tool after calling `self.env.step` so we can make sure the rewrite "worked". We could re-generate the instructions by calling `info['instructions']=self.env.instructions` (which will generate the tools instructions) after adding the `pdb_tool` back to `self.env.tools`.

mormio: Much better idea, thank you! Updated

froggy/agents/zero_shot.py
```diff
@@ -115,3 +115,78 @@ def build_question_prompt(self):
...
         messages.append({"role": "user", "content": question})
         return messages
+
+
+class AgentZeroShot_PdbAfterRewrites(AgentZeroShot):
+    name: str = "zero shot pdb after rewrites"
+
+    def __init__(self, config_dict, env, verbose=False, _uuid=None):
+        super().__init__(config_dict, env, verbose, _uuid)
+        self.history = HistoryTracker(self.config["memory_size"])
```
MarcCote: The history is now created in the BaseAgent.
```suggestion
```

PR #17, opened 2024-11-25
Test tools

chisingh:


.github/workflows/tests.yml
```diff
@@ -33,4 +33,9 @@ jobs:
       - name: Test with pytest
         run: |
           pytest -n 4 -vv --cov=.
+      - name: Generate coverage report
+        run: |
           coverage report -m
+      - name: Check pytest coverage
```
matheper: Maybe replace `Test with pytest` with this `Check pytest coverage` (adding `cov-fail-under` param) because tests are running twice, one for each step.

chisingh: The problem with that approach was that report would not be generated if threshold was not exceeded.

froggy/envs/aider.py
```diff
@@ -110,7 +106,7 @@ def make_froggyignore(self, directory: str, include_gitignore: bool = True):
                 froggyignore_contents += "\n"
                 froggyignore_contents += gitignore_content
 
-        with open(directory / ".froggyignore", "w") as f:
+        with open(directory + "/.froggyignore", "w") as f:
```
matheper: Considering the code from `load_dataset`, where this function is called, I believe `directory` is of type `pathlib.Path` which implement the `/` operator for concatenating directories. In the tests you could also pass a `Path` instead of `str` so you don't need to change the concatenation method here (just update the `directory` type hint, which should be `Path`, not `str`)

froggy/envs/swe_bench.py
```diff
@@ -114,5 +110,5 @@ def make_froggyignore(self, local_repo_path, include_gitignore: bool = True):
                 froggyignore_contents += "\n"
                 froggyignore_contents += gitignore_content
 
-        with open(local_repo_path / ".froggyignore", "w") as f:
+        with open(local_repo_path + "/.froggyignore", "w") as f:
```
matheper: I think the same comment from aider.py applies here

froggy/envs/aider.py
matheper: This code could also be simplified
from: `with open(pjoin(directory, ".gitignore"), "r") as f:`
to: `with open(directory / ".gitignore", "r") as f:`

froggy/tools/patchers.py
```diff
@@ -58,7 +58,7 @@ def use(self, patch):
     def _apply_unified_diff(self, code_lines, patch_lines):
         import re
 
-        patched_lines = code_lines.copy()
+        patched_lines = code_lines.split().copy()
```
matheper: It seems this method already expects the `code_lines` to be a list of strings, it's called at line 41:
```
        patched_code = self._apply_unified_diff(
            code, patch_lines.splitlines(keepends=True)
        )
```

froggy/tests/test_aider.py
```diff
@@ -0,0 +1,76 @@
...
+from froggy.utils import cleanup_pytest_output, extract_max_score_from_pytest_output, extract_reward_from_pytest_output
+
+@pytest.fixture
+@patch('subprocess.run')
+@patch('os.path.exists', return_value=False)
+@patch('pathlib.Path.iterdir')
+@patch('pathlib.Path.read_text', return_value="Test instructions")
+@patch('os.listdir', return_value=[".gitignore"])
+@patch('builtins.open', new_callable=mock_open)
+def aider_env(mock_open, mock_listdir, mock_read_text, mock_iterdir, mock_exists, mock_run):
```
matheper, line 14: Do you think it would be possible to create a dummy aider repo instead of mocking all the os operations? It would be nice if the code runs on a small dummy repo. It's probably more complicated here, but I'm thinking of something like the `setup_test_repo` from `froggy/tools/tests/test_pdb.py`

chisingh: Makes sense. May be we can add some integration tests later after unit tests are done.

.github/workflows/tests.yml
matheper: It seems we could run the 3 steps together with:
 
`pytest -n 4 -vv --cov=. --cov-report term-missing --cov-fail-under=90`
 
On the other hand, we would not know if the step failed because the tests didn't pass, or the coverage dropped. I'm happy with both options.

PR #19, opened 2024-11-28
pass entrypoint properly for aider

sordonia:


froggy/envs/env.py
```diff
@@ -90,6 +92,8 @@ def setup_workspace(
         entrypoint: str,
         readonly_patterns: list[str] = None,
     ):
+        assert entrypoint.split()[0] == "python", "Only support python entrypoint for now."
+
```
MarcCote: I know it is counter-intuitive but given the current workflow, we only instantiate one RepoEnv and use setup_workspace to switch between problem which is the function that uses entrypoint, so the check should go in there.

MarcCote: Oops I misread the Github diff :D. Nvm

froggy/envs/env.py
```diff
@@ -122,12 +124,11 @@ def setup_workspace(
         self.editable_files = [
             p for p in self.all_files if not self.is_readonly(self.working_dir / p)
         ]
+        assert entrypoint.split()[0] == "python", "Only support python entrypoint for now."
```
matheper: SWE need setup_workspace to set the new entrypoint. There's no way to know the entrypoint until you know the example you are trying to solve.

MarcCote: Oh and also I guess that assert won't work with with swe-bench :/

PR #21, opened 2024-11-29
support for threads in run

sordonia:
- supports threads in run
- don't use print, use logger
- wait for #19  and #20

### DIFF
--- a/froggy/terminal.py
+++ b/froggy/terminal.py
@@ -5,9 +5,7 @@ import io
 import logging
 import os
 import pty
-import random
 import shlex
-import string
 import subprocess
 import tempfile
 import termios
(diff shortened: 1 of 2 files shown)


scripts/run.py
```diff
@@ -85,11 +110,28 @@ def main():
...
+def main():
+    config, args = load_config()
+    available_agents = list(config.keys())
+    assert (
+        args.agent in available_agents
+    ), f"Invalid agent. Available agents: {available_agents}"
+
+    config = config[args.agent]
+
+    env = create_env(args, config)
```
MarcCote, line 134: Is that still useful?

MarcCote: Nvm, it is used to list the problem via env.dataset.keys(). Feels like that list of problem should be hardcoded for the benchmark. What do you think?

MarcCote: Nvm. I think that's okay to do it this way.

PR #22, opened 2024-12-04
Text generation using vllm

matheper:
TODO:
Reloading model (save checkpoint, `vllm.free_memory()`, then `vllm.instantiate_llm(checkpoint_path)`

### DIFF
--- a/smol.py
+++ b/smol.py
@@ -171,10 +171,10 @@ class PPOLLM:
             messages, add_generation_prompt=True, tokenize=False,
         )
 
-        response = self.vllm.generate(messages_chat)
+        response = self.vllm.generate(messages_chat)[0]
         token_usage = {
             "prompt": self.token_counter(messages=messages),  # or messages_chat?
-            "response": self.token_counter(text=response[0]),
+            "response": self.token_counter(text=response),
         }
         return response, token_usage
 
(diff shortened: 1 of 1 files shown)


smol.py
```diff
@@ -28,7 +39,7 @@
 batch_size = 10
 
 config = {
-    "output_path": "/root/outputs",
+    "output_path": "/tmp/outputs",
```
chisingh, line 42: should this folder name be taken from base_dir argument?
```suggestion
    "output_path": "/tmp/outputs",
```

smol.py
```diff
@@ -152,15 +166,29 @@ def clear_tensors(self):
...
 
-    def __call__(self, messages, *args, **kwargs):
+    def vllm_generate(self, messages):
+        messages_chat = self.tokenizer.apply_chat_template(
+            messages, add_generation_prompt=True, tokenize=False,
+        )
 
+        response = self.vllm.generate(messages_chat)[0]
+        token_usage = {
+            "prompt": self.token_counter(messages=messages),  # or messages_chat?
```
chisingh, line 176: It makes more sense to count the tokens in actual prompt from messages_chat.

smol.py
matheper: Yes. This setting is overridden below: `config["output_path"] = str(base_dir / "outputs")`
I kept it here because I don't want to change @charbel08 's code too much.

smol.py
matheper: I think so too, that's why I added the comment. What you think @charbel08?

PR #25, opened 2024-12-12
swe-bench integration

matheper:
I suggest we review this PR as is and implement anything missing in a separate one.

TODO (another PR):
- How to run pdb with repos that don't use pytest
- Using `FAIL_TO_PASS` and `PASS_TO_PASS`? How to use that when tests are not pytest?
- Support running tasks in parallel (either instantiating multiple SWEBenchEnv objects or using a single one, but creating separate workspaces per task).
- Cache built images if running from a DockerTerminal

### DIFF
--- a/froggy/utils.py
+++ b/froggy/utils.py
@@ -187,6 +187,7 @@ def load_config():
         "--debug", action="store_true", help="Before sending action to the environment."
     )
     parser.add_argument("-v", "--verbose", action="store_true", help="Verbose mode")
+    parser.add_argument("-vv", "--very-verbose", action="store_true", help="Set logging level to DEBUG")
     parser.add_argument(
         "-p",
         "--params",
--- a/scripts/run.py
+++ b/scripts/run.py
@@ -104,11 +104,16 @@ def create_agent(args, config):
 
     if args.verbose:
         agent.llm.verbose = True
+
     return agent
 
 
 def main():
     config, args = load_config()
+    if args.very_verbose:
+        args.verbose = True
+        logger.setLevel(logging.DEBUG)
+
     available_agents = list(config.keys())
     assert (
         args.agent in available_agents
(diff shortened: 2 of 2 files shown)


froggy/envs/swe_bench.py
```diff
@@ -16,12 +18,22 @@
...
+        dataset_id: str = "princeton-nlp/SWE-bench_Verified",
+        split: str = "test",
+        base_image: str = "python:3.12",
+        **kwargs,
+    ):
         super().__init__(**kwargs)
+        self.dataset_id = dataset_id
+        self.split = split
+        self.swe_bench_base_image = base_image
+        self.swe_bench_repo_paths = Path(pjoin(tempfile.gettempdir(), "swe-bench"))
```
chisingh: Better to have this inside user's home directory or the workspace folder itself.

froggy/envs/swe_bench.py
```diff
@@ -109,10 +137,171 @@ def make_froggyignore(self, local_repo_path, include_gitignore: bool = True):
...
                 gitignore_content = f.read()
                 froggyignore_contents += "\n"
                 froggyignore_contents += gitignore_content
 
         with open(local_repo_path / ".froggyignore", "w") as f:
             f.write(froggyignore_contents)
+
+    def run_command_with_raise(self, command):
+        if "apt-get" in command:
+            command = command.replace("apt-get", "sudo apt-get")
```
chisingh: what if the command already begins with sudo?

matheper: Good question. There's no command that includes sudo or other commands that require sudo in swe-bench. We can see all of them in `swebench.harness.constants.MAP_REPO_VERSION_TO_SPECS`. 
I don't like concatenating sudo to the command, but I couldn't find another way to do that. If you have any suggestion, I'm happy to implement them.

chisingh: Maybe add a check whether the command already doesn't begin with a sudo and contains apt-get.

froggy/envs/swe_bench.py
matheper: Makes sense! Changed to user's home directory.

PR #31, opened 2025-01-09
Integrate function call in API

Kim-Minseon:
- Update "comparison_aider.py" to draw multiple figures for analysis
- Update "llm_api.py" and "zero_shot.py" to use function call (tool call) inside the api
- Small update to use uuid as an argument

### DIFF
--- a/analysis/README.md
+++ b/analysis/README.md
@@ -1,0 +1,66 @@
+# Analysis Folder
+
+This folder contains scripts for analyzing data after runing the agent through the Froggy. Below is a brief description of each script:
+
+## Scripts
+
+### gather_results_aider.py
+> This script gathers success rate results from all problems in the aider bench.
+
+### comparisons_aider.py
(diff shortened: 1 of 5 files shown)


scripts/run.py
```diff
@@ -88,8 +88,7 @@ def create_agent(args, config):
     match args.agent:
         case "zero_shot":
             from froggy.agents import AgentZeroShot
-
-            agent = AgentZeroShot(config, env, verbose=args.verbose)
+            agent = AgentZeroShot(config, env, verbose=args.verbose, _uuid=args.uuid)
```
MarcCote, line 89: I think we need to rethink this uuid thing.

froggy/utils.py
```diff
@@ -183,6 +183,7 @@ def load_config():
     parser = argparse.ArgumentParser()
     parser.add_argument("config_file", help="path to config file")
     parser.add_argument("--agent", help="zero_shot, cot, tadpole", default="zero_shot")
+    parser.add_argument("--uuid", help="uuid for agent", default=None)
```
MarcCote, line 186: What configs should come from the YAML file and which ones from the command line?
@xingdi-eric-yuan

PR #35, opened 2025-01-28
Use logging instead of print. Add nice progress visualization.

MarcCote:
In addition to switch to logging (file and stdout) instead of print, and adding a nice progress visualization (see image), this pull request also includes the following changes:
- Remove the explicit `_uuid` params in the `AgentBase` (other agents) constructor and use the `uuid` field from the `config` dictionary instead.
- Add typing
- Move the `is_rate_limit_error` inside the `LLM` class (needed for logging).
- Set a timeout of 1 second to stop container (instead of the default which is 10 secs).
- Provide an example in `zero_shot.py` of how we could log progress instead of using tqdm. 

image

### DIFF
--- a/froggy/agents/zero_shot.py
+++ b/froggy/agents/zero_shot.py
@@ -9,9 +9,6 @@ from froggy.utils import unescape
 class AgentZeroShot(AgentBase):
     name: str = "zero shot"
 
-    def __init__(self, config_dict, env, verbose=False, _uuid=None):
-        super().__init__(config_dict, env, verbose, _uuid)
-
     def build_question_prompt(self):
         messages = []
         question = "Based on the instruction, the current code, the last execution output, and the history information, "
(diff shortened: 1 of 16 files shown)


froggy/agents/zero_shot.py
```diff
@@ -38,14 +35,20 @@ def run(self, task_name=None, debug=False):
...
-            total=self.config["max_steps"],
-            desc=f"Debugging inside {self.env.working_dir} - Task: {task_name}",
-            leave=True,
-        )
+        # pbar = tqdm(
+        #     total=self.config["max_steps"],
+        #     desc=f"Debugging inside {self.env.working_dir} - Task: {task_name}",
+        #     leave=True,
+        # )
+        self.logger.progress.update(self.logger.task_id, total=self.config["max_steps"])
```
MarcCote: To discuss, @matheper

scripts/config_aider.yaml
matheper, line 1: Do we merge these changes?

froggy/utils.py
```diff
@@ -186,6 +186,24 @@ def load_config():
         "--debug", action="store_true", help="Before sending action to the environment."
     )
     parser.add_argument("-v", "--verbose", action="store_true", help="Verbose mode")
+    parser.add_argument(
```
matheper: are the levels debug (-vv), info (-v), and warning if no flag?

MarcCote: I think with no flag it is Info

MarcCote: I don't think there any difference between -v and -vv now. :D

scripts/config_aider.yaml
MarcCote: Oops no

froggy/utils.py
MarcCote: Done

scripts/config_aider.yaml
MarcCote: REmoved

matheper (approved):

PR #36, opened 2025-01-29
Clean Tools

xingdi-eric-yuan:
1) remove whole / udiff patchers
2) remove repeating descriptions from tools
3) improve the reasoning tool. In the previous version, when `allow_chain_action is True`, the `use_with_chaining()` function calls `self.env.step()` and only returns the new observation, but does not update the info dict. Now it should work properly.

### DIFF
--- a/froggy/envs/env.py
+++ b/froggy/envs/env.py
@@ -339,10 +339,8 @@ class RepoEnv(TooledEnv):
                     and self.get_tool(triggered_tool.name).infos_cache is not None
                 ):
                     # use done, score and info from the tool that was executed after reasoning
-                    self.done = copy.copy(self.get_tool(triggered_tool.name).done_cache)
-                    self.score = copy.copy(
-                        self.get_tool(triggered_tool.name).score_cache
-                    )
+                    self.done = self.get_tool(triggered_tool.name).done_cache
+                    self.score = self.get_tool(triggered_tool.name).score_cache
                     self.infos = copy.copy(
                         self.get_tool(triggered_tool.name).infos_cache
                     )
--- a/froggy/tools/reasoning.py
+++ b/froggy/tools/reasoning.py
@@ -93,9 +93,9 @@ You break down complex problems into smaller parts and reason through them step 
                 ["Error while executing the action after reasoning.", next_obs]
             )
         self.success_chain_action = True
-        self.done_cache = copy.copy(next_done)
+        self.done_cache = next_done
+        self.score_cache = next_score
         self.infos_cache = copy.copy(next_infos)
-        self.score_cache = copy.copy(next_score)
         return "\n".join(
             [
                 "Reasoning:",


froggy/tools/reasoning.py
```diff
@@ -5,38 +5,46 @@
...
+        else:
+            return "```reasoning ... ```"
+
     @property
     def examples(self):
         if self.allow_chain_action:
             ex = [
-                "<reasoning> The execution trace points to line 43 in main.py, so I'll place a breakpoint there.</reasoning> ```pdb b 43",
-                "<reasoning> There's a shape mismatch that corresponds to a matrix transpose, so I'll rewrite the function to account for the transpose. </reasoning> ```rewrite ....",
+                "```The execution trace points to line 43 in main.py, so I'll place a breakpoint there.</reasoning> ```pdb b 43``` ```",
```
sordonia: watch out for <\/reasoning>

froggy/tools/reasoning.py
```diff
@@ -5,38 +5,46 @@
...
+                "```There's a shape mismatch that corresponds to a matrix transpose, so I'll rewrite the function to account for the transpose. </reasoning> ```rewrite 10 <c>    m = m.transpose()</c>``` ```",
```
sordonia: same here, is this wanted?

froggy/tools/listdir.py
```diff
@@ -8,7 +8,6 @@
 class ListdirTool(EnvironmentTool):
     name: str = "listdir"
     action: str = "```listdir"
-    description: str = "List the directory tree at a given subdirectory."
```
matheper, line 11: Any reason to remove the ListdirTool description?

froggy/tools/reasoning.py
```diff
@@ -5,38 +5,46 @@
...
             ]
         else:
             ex = [
-                "<reasoning> The execution trace points to line 43 in main.py, so I'll place a breakpoint there.</reasoning> ",
-                "<reasoning> There's a shape mismatch that corresponds to a matrix transpose, so I'll rewrite the function to account for the transpose. </reasoning>",
+                "```The execution trace points to line 43 in main.py, so I'll place a breakpoint there.```",
```
sordonia: Missing "```reasoning" at the beginning?

froggy/tools/reasoning.py
```diff
@@ -5,38 +5,46 @@
...
+                "```There's a shape mismatch that corresponds to a matrix transpose, so I'll rewrite the function to account for the transpose.```",
```
sordonia: "```reasoning"

froggy/tools/reasoning.py
```diff
@@ -67,16 +75,27 @@ def use_with_chaining(self, action):
...
+        except:
+            return "SyntaxError: invalid syntax."
         return "Reasoning text acknowledged."
 
-    def remove_reasoning(self, action):
-        items = action.split("</reasoning>")
-        return " ".join(items[1:]).lstrip()
+    def split_reasoning(self, action):
+        content = action.split(self.action)[1].rsplit("```", 1)[0].strip()
+        reasoning, next_action = content.split("</reasoning>", 1)
```
sordonia, line 119: "<\/reasoning>" still here, is this desirable?

froggy/tools/reasoning.py
xingdi-eric-yuan: Good catch!

froggy/tools/reasoning.py
xingdi-eric-yuan: Yes, the reasoning tool was designed to preface any reasoning text before generating an actual action, "<//reasoning>" is used to split the reasoning text and the action.

froggy/tools/reasoning.py
```diff
@@ -5,38 +5,46 @@
...
+                "```There's a shape mismatch that corresponds to a matrix transpose, so I'll rewrite the function to account for the transpose.```",
             ]
         return ex
 
     @property
     def description(self):
         if self.allow_chain_action:
             desc = f"""You may explicitly reason about the current state and the best course of action before executing. You follow a particular reasoning style:
-You break down complex problems into smaller parts and reason through them step by step, arriving at the best next action before then executing it. You should follow your reasoning with your next action. """
+You break down complex problems into smaller parts and reason through them step by step, arriving at the best next action before then executing it. You should follow your reasoning with your next action. The next action should be a valid action that correctly follows the syntax can be executed in the current environment. The next action cannot be another reasoning action."""
```
matheper: `syntax and can be...`?:

`The next action should be a valid action that correctly follows the syntax and can be executed in the current environment.`

xingdi-eric-yuan: how about 
`valid action, which follows the syntax rules defined by the tools available in the current environment` ?

froggy/tools/reasoning.py
```diff
@@ -67,16 +75,27 @@ def use_with_chaining(self, action):
...
         # now execute the next action
-        next_action_obs = self.environment.step(remaining_action)
+        next_action_obs = self.environment.step(next_action)
         if next_action_obs == f"Invalid action: {action}.":
             next_action_obs == f"You must provide a valid action after your reasoning. Found invalid action: {action}."
-        return f"Reasoning text acknowledged. {next_action_obs[0]}"
+        return f"Reasoning text acknowledged.\n{next_action_obs[0]}"
 
-    def use_without_chaining(self):
+    def use_without_chaining(self, action):
```
matheper, line 110: Could you write a test for these three methods: `use_with_chaining`, `use_without_chaining`, `split_reasoning`?

xingdi-eric-yuan: Will do

xingdi-eric-yuan: Done

froggy/envs/env.py
```diff
@@ -330,6 +332,24 @@ def step(self, action: str):
                     current_frame_file = self.get_tool("pdb").current_frame_file
                     if current_frame_file in self.all_files:
                         self.load_current_file(self.get_tool("pdb").current_frame_file)
+            elif isinstance(triggered_tool, ReasoningTool):
```
MarcCote, line 335: Maybe not in this PR, but the whole tool system needs to be redesigned. It should probably be event-based and the env.py shouldn't be aware of what tools exist if we truly want to be modular.

xingdi-eric-yuan: Yes I agree.

froggy/envs/env.py
```diff
@@ -330,6 +332,24 @@ def step(self, action: str):
...
+                if (
+                    self.get_tool(triggered_tool.name).success_chain_action is True
+                    and self.get_tool(triggered_tool.name).done_cache is not None
+                    and self.get_tool(triggered_tool.name).infos_cache is not None
+                ):
+                    # use done, score and info from the tool that was executed after reasoning
+                    self.done = copy.copy(self.get_tool(triggered_tool.name).done_cache)
+                    self.score = copy.copy(
```
MarcCote: why copy.copy here? This is a int, no?

froggy/envs/env.py
```diff
@@ -330,6 +332,24 @@ def step(self, action: str):
...
                     if current_frame_file in self.all_files:
                         self.load_current_file(self.get_tool("pdb").current_frame_file)
+            elif isinstance(triggered_tool, ReasoningTool):
+                if (
+                    self.get_tool(triggered_tool.name).success_chain_action is True
+                    and self.get_tool(triggered_tool.name).done_cache is not None
+                    and self.get_tool(triggered_tool.name).infos_cache is not None
+                ):
+                    # use done, score and info from the tool that was executed after reasoning
+                    self.done = copy.copy(self.get_tool(triggered_tool.name).done_cache)
```
MarcCote: No need for copy, done is a bool

froggy/tools/reasoning.py
```diff
@@ -67,16 +69,53 @@ def use_with_chaining(self, action):
...
+                ]
+            )
+        if next_obs == f"Invalid action: {action}." or next_obs.startswith(
+            "Error while using tool"
+        ):
+            return "\n".join(
+                ["Error while executing the action after reasoning.", next_obs]
+            )
+        self.success_chain_action = True
+        self.done_cache = copy.copy(next_done)
```
MarcCote: Same

PR #39, opened 2025-01-31
Refac tests

matheper:
- move tests to their respective folders
- fixed tests leaving empty folders at the root dir
- refac tests using pytest style

### DIFF
--- a/scripts/config_swebench.yaml
+++ b/scripts/config_swebench.yaml
@@ -1,6 +1,6 @@
 zero_shot:
     # Environment configs
-    output_path: "exps"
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
+output_swe_bench/
```
MarcCote: I think we should change the default output path to be `exps/` and then add `exps/` in the .gitignore

MarcCote: Can be done in a next PR. Let's wait for the tests and merge.

matheper: Done

PR #40, opened 2025-02-03
Add .froggyignore and .froggyreadonly

MarcCote:


pyproject.toml
```diff
@@ -19,8 +19,9 @@ classifiers = [
 [tool.setuptools.dynamic]
 dependencies = {file = ["requirements.txt"]}
 
-[tool.setuptools.packages]
-find = {}
+[tool.setuptools.packages.find]
```
matheper, line 22: Nice! Thanks for adding this! 😄

froggy/utils.py
```diff
@@ -184,7 +225,12 @@ def load_config():
     parser.add_argument("config_file", help="path to config file")
     parser.add_argument("--agent", help="zero_shot, cot, tadpole", default="zero_shot")
     parser.add_argument(
-        "--debug", action="store_true", help="Before sending action to the environment."
+        "--debug",
+        action="store_true",
+        help="Break before sending action to the environment.",
+    )
+    parser.add_argument(
+        "--editable", action="store_true", help="Break before sending prompt to agent."
```
matheper: What is the help message for --editable?

MarcCote: Oops that was change from my debug branch. I wanted to have the oppotunity to break before sending the prompt to the LLM to amend it.

MarcCote: Gone

PR #50, opened 2025-02-10
Add Dataclasses and RepoEnv Info refac

matheper:
- Create some dataclasses to keep track of the environment information (`EnvInfo`) and LLM responses (`LLMResponse` and `TokenUsage`)
- Methods `step` and `reset` always return an `EnvInfo` object instead of tuples
- LLM always returns an `LLMResponse`
- Move token usage tracking to `HistoryTracker`
- Combine `HistoryTracker.save_prompt_response_pairs` into `HistoryTracker.step`
- Remove Random LLM
- Test tool and toolbox

### DIFF
--- a/example_agent/utils.py
+++ b/example_agent/utils.py
@@ -19,7 +19,7 @@ class HistoryTracker:
     def step(self, new_info: EnvInfo, llm_response: LLMResponse | None = None) -> None:
         """llm_response can be None since the initial state does not have prompt and response"""
         self.memory.append(copy.deepcopy(new_info))
-        self.prompt_response_pairs.append(copy.deepcopy(llm_response))
+        self.prompt_response_pairs.append([copy.deepcopy(llm_response)])
 
     def get(self):
         # return the history_steps latest steps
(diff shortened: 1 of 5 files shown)


example_agent/utils.py
```diff
@@ -41,28 +44,29 @@ def json(self, game_step=None, include_prompt_response_pairs=False):
...
+                "action": self.memory[game_step].action,
+                "obs": self.memory[game_step].obs,
             }
-            if include_prompt_response_pairs:
-                json_out["prompt_response_pairs"] = self.prompt_response_pairs[
-                    game_step
-                ]
+            prp = self.prompt_response_pairs[game_step]
+            if prp and include_prompt_response_pairs:
+                json_out["prompt_response_pairs"] = self._format_prompt_response_pairs(
```
matheper: I'm changing the format of the `prompt_response_pairs` when the prompt is a list of messages, see `_format_prompt_response_pairs` and `tests/test_agents.py:193`. I'm not sure I understand the previous logic when there are multiple message turns. Now the messages are concatenated into one prompt.

xingdi-eric-yuan: This was for the cases where there are multiple steps of LLM calls in a single env.step. For instance, in some CoT settings, we first call the LLM to generate the reasoning string, then conditioned on this string, we ask the LLM again to generate an action. I believe it's better to keep the list instead of concat, because list can be better presented when doing json.dumps when creating prompts.

froggy/envs/env.py
```diff
@@ -207,26 +226,27 @@ def reset(self, *, seed=None, options: dict = None, restore_code=True):
...
-            "dbg_obs": self.dbg_obs if hasattr(self, "dbg_obs") else "",
-            "last_run_obs": self.last_run_obs,
-            "dir_tree": self.display_files(),
-            "current_breakpoints": (
+        self.infos = EnvInfo(
+            obs=self.obs,
+            dbg_obs=self.dbg_obs if hasattr(self, "dbg_obs") else "",
+            last_run_obs=self.last_run_obs,
+            dir_tree=self.display_files(),
+            current_breakpoints=(
```
xingdi-eric-yuan, line 234: Not sure in this PR or the next one, `current_breakpoints` should be part of the `pdb` tool state, not env state.

matheper: I think we can make that in the next PR. I'm probably moving the pdb_obs to the tool as well

example_agent/utils.py
matheper: I removed the string format, but moved the token_usage inside `prompt_response_pair` since there's usage for each of the llm calls

PR #58, opened 2025-02-14
EventHooks - Tools refac

matheper:
Refac introducing events to coordinate how env and tools interact with each other

### DIFF
--- a/froggy/envs/tests/test_env.py
+++ b/froggy/envs/tests/test_env.py
@@ -84,7 +84,7 @@ def test_tool_names(env_mock):
     tool2.name = "tool2"
     env_mock.add_tool(tool1)
     env_mock.add_tool(tool2)
-    assert env_mock.tool_names == "tool1, tool2"
+    assert env_mock.tool_names == "```tool1```, ```tool2```"
 
 
 def test_tool_instructions(env_mock):


froggy/envs/env.py
```diff
@@ -37,9 +43,56 @@ class EnvInfo:
...
+        self.event_listeners[event].append(tool)
+
+    def unsubscribe(self, event: Event, tool):
+        self.event_listeners[event].remove(tool)
+
+    def notify(self, event: Event, source=None, **kwargs):
+        observations = []
+        for tool in self.event_listeners[event]:
+            if tool == source:
+                continue  # skip the source tool to avoid infinite loop
```
MarcCote, line 60: Should we detect cycles? ToolA -> ToolB -> ToolA -> ...

froggy/envs/env.py
```diff
@@ -37,9 +43,56 @@ class EnvInfo:
...
+
+    @classmethod
+    def list(cls):
+        """Returns list of event names as strings"""
+        return [event.value for event in cls]
+
+
+class EventHooks:
+    def __init__(self):
+        self.events = list(Event)
```
MarcCote: Why passing `Event` to `list` this?

froggy/envs/env.py
MarcCote: Actually, `tool == source` might be fine. A tool could listen to its own event. What we don't want is an event handler that triggers the same event.

froggy/envs/env.py
```diff
@@ -221,22 +277,15 @@ def reset(self, *, seed=None, options: dict = None, restore_code=True):
         self.run()
```
MarcCote: TODO: move to EvalTool with on_env_reset

froggy/envs/env.py
```diff
@@ -350,33 +417,18 @@ def step(self, action: str):
...
 
+        observations = []
         self.obs = f"Invalid action: {action}."
         if triggered_tools:
             triggered_tool = triggered_tools[0]
             try:
-                self.obs = triggered_tool.use(action)
+                observations += triggered_tool.use(action)
+                self.obs = ""
+                # self.obs = f"Success using tool {triggered_tool.name}"
```
MarcCote: Remove

froggy/envs/env.py
```diff
@@ -413,3 +461,23 @@ def step(self, action: str):
...
+    def handle_event(self, event: Event, source=None, **kwargs):
+        observations = []
+        obs = ""
+        if event in [Event.REWRITE_SUCCESS, Event.REWRITE_FAIL]:
+            self.rewrite_counter += 1
+
+        if event == Event.REWRITE_FAIL and self.run_on_rewrite:
+            # obs += "\nNew code has been run."
+            run_obs, _done = self.run()  # remove _done from run
+            # observations += [{"environment_run": run_obs}]  # move dict inside run method
```
MarcCote: TODO: move to EvalTool

MarcCote: Need to decide on where the _done will live though.

froggy/tools/pdb.py
```diff
@@ -86,6 +86,18 @@ def start_pdb(self, pdb_cmd: str = None) -> str:
...
 
+    def on_env_reset(self):
+        return [{self.name: self.start_pdb()}]  # start or restart?
+
+    def on_rewrite(self):
+        obs =  self.restart_pdb()
+        obs = "\nDebugging terminal started:\n" f"{obs}\n"
+        return [{self.name: obs}]
+
+    def on_rewrite_success(self, rewrite_file, rewrite_head, rewrite_tail, new_code_length, **kwargs):
```
MarcCote: We could simplify the params name a bit.
```suggestion
    def on_rewrite_success(self, file, head, tail, length, **kwargs):
```

froggy/tools/tool.py
```diff
@@ -22,5 +22,8 @@ def is_triggered(self, action):
         return action.startswith(self.action)
 
     @abstractmethod
-    def use(self, action, environment):
-        raise NotImplementedError("use method must be implemented in the subclass.")
+    def use(self, action, environment) -> list[dict]:
+        pass
+
+    def trigger_event(self, event: "Event", **kwargs):  # TODO: import Event from froggy.envs.env
```
MarcCote: some Todo

scripts/run.py
```diff
@@ -83,6 +84,9 @@ def create_env(config: dict, logger: FroggyLogger):
 
         tool_instantiated = Toolbox.get_tool(tool, **kwargs)
         env.add_tool(tool_instantiated)
+        if tool == "pdb":
+            env.event_hooks.subscribe(Event.ENV_RESET, tool_instantiated)
+            env.event_hooks.subscribe(Event.REWRITE_SUCCESS, tool_instantiated)
```
MarcCote: Shouldn't that happen in the tool? Maybe when registering it to the env.

froggy/envs/env.py
matheper: I think cycles are fine. If a different tool sends a new event, the first (tool) coud react to that, if needed.
For your second point, the tool doesn't need the event system to perform that action, it can just perform their thing before sending the event. This way we can avoid the cycles by just skipping tool == source. Does that make sense?

froggy/envs/env.py
matheper: Done!

froggy/envs/env.py
matheper: Partially done. Using EvalTool now, but it's still calling RepoEnv.run, which set RepoEnv.done.

scripts/run.py
matheper: I think this should go to the config, so you are free to design your environment/tools as you wish. Another option I played with is to auto-subscribe to the events the tools implement hooks for, for example:
```python
class EnvironmentTool(ABC):
...
    # Auto-subscribe to events that have handlers
    for event in Event:
        if hasattr(self, event.handler_name):
            environment.event_hooks.subscribe(event, self)
```

froggy/tools/pdb.py
matheper: Done

froggy/tools/tool.py
matheper: Yes, I'm planning to move the dataclasses somewhere else later.

froggy/envs/env.py
matheper: Good question. I'm not sure which one is better:
```
In [1]: list(Event)
Out[1]: 
[<Event.ENV_START: 'env_start'>,
 <Event.ENV_RESET: 'env_reset'>,
 <Event.ENV_STEP: 'env_step'>,
 <Event.FILE_CHANGE: 'file_change'>,
 <Event.REWRITE_SUCCESS: 'rewrite_success'>,
 <Event.REWRITE_FAIL: 'rewrite_fail'>,
 <Event.SWITCH_CONTEXT: 'switch_context'>]

In [2]: Event
Out[2]: <enum 'Event'>

```

matheper: Probably the enum. We can still do `event in Event`

froggy/envs/swe_bench.py
```diff
@@ -314,17 +313,8 @@ def _extract_score(self, obs):
         assert score <= self.max_score
         return score
 
-    def step(self, action: str):
```
matheper, line 317: No special logic here, using `RepoEnv.step` instead. Score related code was moved to `eval`, previously called `run`.

froggy/envs/swe_bench.py
```diff
@@ -189,6 +189,15 @@ def patch(self):
         patch = result.stdout.replace(str(self.working_dir), str(self.path))
         return patch
 
+    def eval(self, **kwargs):
+        success, output = self.terminal.run(self.entrypoint, timeout=self.run_timeout)
+        # TODO: probably needed cleanup specific to each SWE-Bench repo. Does it?
```
matheper: Do we need any cleanup @MarcCote ?

froggy/envs/mini_nightmare.py
```diff
@@ -45,15 +53,7 @@ def reset(self, *, seed=None, options: dict = None):
...
-        infos.instructions = self.instructions
-        infos.last_run_obs = utils.cleanup_pytest_output(infos.last_run_obs)
-
-        self.max_score = utils.extract_max_score_from_pytest_output(infos.last_run_obs)
-        infos.max_score = self.max_score
-        self.score = utils.extract_reward_from_pytest_output(infos.last_run_obs)
-        self.done = self.score == self.max_score
-        infos.score = self.score
-        infos.done = self.done
+        infos.instructions = self.instructions  # TODO: is this needed?
```
matheper: Any reason this was needed?

example_agent/example_agent.py
```diff
@@ -248,7 +253,9 @@ def run(self, task_name=None, debug=False):
                 self.env.add_tool(pdb_tool)
                 self.env.tools["pdb"].start_pdb()
                 info.instructions = self.env.instructions
-                info.obs += "\nThe pdb tool has been added."
+                info.step_observation.observation += (
+                    "\nThe pdb tool has been added."  # Does this affect anything?
```
matheper: Do we need this?

froggy/envs/mini_nightmare.py
xingdi-eric-yuan: I don't think it's necessary.

example_agent/example_agent.py
xingdi-eric-yuan: I think it's fine to remove this, because agent will suddenly see pdb instructions being added into its system prompt. However, let's check and make sure my last sentence is true 🤣

froggy/envs/aider.py
```diff
@@ -23,32 +23,30 @@ def __init__(self, **kwargs):
...
         directory = self.current_sample["base_directory"]
         self.setup_workspace(directory, entrypoint="python -m pytest -s .")
         infos = super().reset()
-        infos.instructions = self.instructions
-        infos.last_run_obs = utils.cleanup_pytest_output(infos.last_run_obs)
-
-        self.max_score = utils.extract_max_score_from_pytest_output(infos.last_run_obs)
-        infos.max_score = self.max_score
-        infos.score = utils.extract_reward_from_pytest_output(infos.last_run_obs)
+        infos.instructions = self.instructions  # TODO: is this needed?
```
xingdi-eric-yuan: maybe remove

froggy/envs/env.py
```diff
@@ -210,39 +245,40 @@ def restore(self, *filepaths):
 
             shutil.copy2(self.path / filepath, self.working_dir / filepath)
 
-    def reset(self, *, seed=None, options: dict = None, restore_code=True):
+    def reset(
+        self,
+        *,
+        seed=None,
```
xingdi-eric-yuan, line 267: Does seed do anything now?

scripts/config_aider.yaml
```diff
@@ -9,7 +9,7 @@ pdb_agent:
         "auto_view_change": True,
         "run_timeout": 10,
     }
-    tools: ["pdb", "view", "rewrite"]
+    tools: ["eval", "pdb", "view", "rewrite"]
```
xingdi-eric-yuan, line 12: Might need the same for `config.yaml` (for custom repo)

matheper: Done!

froggy/envs/env.py
matheper: Was it doing anything before? If not, it's still doing nothing.

froggy/envs/env.py
matheper: Moved score and done logic to RepoEnv.eval, previously called RepoEnv.run

scripts/run.py
matheper: This PR is now implementing the auto-subscribe

froggy/envs/env.py
matheper: It's now calling `RepoEnv.seed(seed)` on reset

froggy/envs/aider.py
```diff
@@ -23,32 +23,29 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.load_dataset()
 
+    def eval(self, **kwargs):
+        # if not self.done:  # Skip evaluation if the task is already solved.
+        success, output = self.terminal.run(self.entrypoint, timeout=self.run_timeout)
+        self.max_score = utils.extract_max_score_from_pytest_output(output)
+        self.score = utils.extract_reward_from_pytest_output(output)
+        self.done = self.score == self.max_score
+        self.last_eval_obs = utils.cleanup_pytest_output(output)
```
xingdi-eric-yuan, line 30: In the previous version, the `infos.last_run_obs` is first being cleaned:
`infos.last_run_obs = utils.cleanup_pytest_output(infos.last_run_obs)`, 
then, `self.score` is computed by
`utils.extract_reward_from_pytest_output(infos.last_run_obs)`. 

Now the ordering seems reversed, would this cause any potential problem?

MarcCote: No I don't think it causes any issues. The cleanup was to remove anything that appears random and that might throw off the LLMs if added to the context.

xingdi-eric-yuan: OK all good then.

example_agent/example_agent.py
matheper: Removed the string concat.

PR #62, opened 2025-02-19
Lite update tool syntax parser

xingdi-eric-yuan:
1. unified parser for action: e.g., 
```text
` ` `pdb b 10` ` `
```
will be parsed as 
```text
tool_name = "pdb"
tool_args = "b 10"
```
Each `tool.use()` will take `tool_args` as input. 
2. Rename patcher -> rewrite, removing sub-tool for simplicity. 
3. Removing the reasoning tool for simplicity.

### DIFF
--- a/froggy/envs/mini_nightmare.py
+++ b/froggy/envs/mini_nightmare.py
@@ -41,7 +41,7 @@ class MiniNightmareEnv(RepoEnv):
         self.setup_workspace(
             directory,
             entrypoint="python -m pytest -s test.py",
-            debug_entrypoint="pytest --pdb -s test.py",
+            debug_entrypoint="python -m pdb -m pytest -s test.py",
         )
 
         infos = super().reset()
--- a/scripts/config.yaml
+++ b/scripts/config.yaml
@@ -4,7 +4,7 @@ pdb_agent:
     env_kwargs: {
         "path": "data/mini_nightmare/config",
         "entrypoint": "python -m pytest -sv test.py",
-        "debug_entrypoint": "pytest --pdb -s test.py",
+        "debug_entrypoint": "python -m pdb -m pytest -s test.py",
         "dir_tree_depth": 1,
         "run_on_rewrite": True,
         "auto_view_change": True,
@@ -38,7 +38,7 @@ rewrite_only:
     env_kwargs: {
         "path": "data/mini_nightmare/config",
         "entrypoint": "python -m pytest -sv test.py",
-        "debug_entrypoint": "pytest --pdb -s test.py",
+        "debug_entrypoint": "python -m pdb -m pytest -s test.py",
         "dir_tree_depth": 1,
         "run_on_rewrite": True,
         "auto_view_change": True,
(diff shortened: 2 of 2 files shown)


froggy/envs/env.py
```diff
@@ -345,20 +350,17 @@ def patch(self):
...
             try:
-                self.obs = triggered_tool.use(action)
+                self.obs = triggered_tool.use(tool_args)
             except:
                 self.obs = f"Error while using tool {triggered_tool.name} with action: \n{action}"
 
-            if isinstance(triggered_tool, CodePatcher):
+            if isinstance(triggered_tool, RewriteTool):
                 self.rewrite_counter += 1
                 if self.get_tool(triggered_tool.name).rewrite_success:
```
sordonia: self.get_tool(triggered_tool.name) can just be triggered_tool?

froggy/tools/pdb.py
```diff
@@ -159,7 +137,7 @@ def use(self, action):
                     # TODO: add output to self.pdb_obs?
                     obs += f"\nlist .\n" + self.interact_with_pdb("l .")
         else:
-            obs = "\n".join([f"Invalid action: {action}", _warning, output])
+            obs = "\n".join([f"Invalid action: ```pdb{tool_args}```", _warning, output])
```
sordonia: maybe change message to Invalid args: tool_args ?

froggy/envs/env.py
xingdi-eric-yuan: done

froggy/tools/pdb.py
xingdi-eric-yuan: done

froggy/envs/mini_nightmare.py
```diff
@@ -38,22 +38,38 @@ def reset(self, *, seed=None, options: dict = None):
         self.current_sample = self.dataset[options["task_name"]]
 
         directory = self.current_sample["base_directory"]
-        self.setup_workspace(directory, entrypoint="python -m pytest -sv test.py")
+        self.setup_workspace(
+            directory,
+            entrypoint="python -m pytest -s test.py",
+            debug_entrypoint="pytest --pdb -s test.py",
```
matheper: I had some issues running `pytest --pdb`, maybe `python -m pdb -m pytest -s test.py` is better.

matheper: I know what the problem was..  `pytest --pdb` is a post-mortem pdb:

`This will invoke the Python debugger on every failure (or KeyboardInterrupt). Often you might only want to do this for the first failing test to understand a certain failure situation:`

https://docs.pytest.org/en/7.1.x/how-to/failures.html#dropping-to-pdb-on-failures

xingdi-eric-yuan: should we also use `python -m pdb -m pytest -s test.py` for aider?

xingdi-eric-yuan: nvm I think by default it will remove `python` and add `python -m pdb`.

PR #63, opened 2025-02-20
Fix PDB hanging

MarcCote:


froggy/logger.py
```diff
@@ -31,6 +32,11 @@ def __init__(
         super().__init__(name)
         self.setLevel(logging.DEBUG)
 
+        # If var env "LOG_LEVEL" is set, use it as the log level.
```
matheper: LOG_LEVEL or FROGGY_DEBUG?

example_agent/llm_api.py
```diff
@@ -324,6 +324,7 @@ def __call__(self, messages, info, *args, **kwargs) -> LLMResponse:
         # Color each role differently.
         print_messages(messages, self.logger)
         available_commands = [t["template"] for t in info.tools.values()]
+        available_commands.append("```rewrite 1:2 <c>  asd\n asd</c>```")
```
matheper: remove?

froggy/logger.py
MarcCote: I think it might be good to have a env var to control the logging directly (especially when running tests). What do you think? If so, it should FROGGY_DEBUG

example_agent/llm_api.py
MarcCote: yep

froggy/logger.py
matheper: I agree, `FROGGY_DEBUG` is better. I was just confused about the comment referring to `LOG_LEVEL`.

froggy/tools/pdb.py
```diff
@@ -26,30 +26,32 @@ def __init__(
         auto_list: bool = True,
     ):
         super().__init__()
-        self.master = None
         self.pdb_obs = ""
```
sordonia, line 29: do we need to store state in the tools e.g. pdb_obs?

froggy/tools/pdb.py
```diff
@@ -26,30 +26,32 @@ def __init__(
...
+
     def interact_with_pdb(self, command, expected_output="(Pdb)"):
         output = self._session.run(
             command, expected_output, timeout=300, no_output_timeout=300
         )
         return output.replace("(Pdb)", "").strip()  # remove the prompt
 
     def close_pdb(self, command="q"):
-        return self._session.run(command, timeout=10)
+        output = self._session.run(command, timeout=10)
```
sordonia: not sure it's worth issuing 'q'

froggy/tools/pdb.py
MarcCote: I'd say yes. The tool should store its state. But here `pdb_obs` is the "state" (at least not for me), it's just the last observation received from the pdb program.

froggy/tools/pdb.py
MarcCote: Me neither. I didn't want to change it because it was like that before. Maybe @xingdi-eric-yuan or @matheper knows?

xingdi-eric-yuan: I think `self._session.close()` is good enough.

froggy/tools/tests/test_pdb.py
```diff
@@ -313,3 +280,19 @@ def test_get_current_frame_file(mock_interact_with_pdb, tmp_path, setup_pdb_repo
     )
     pdb_tool.get_current_frame_file()
     assert str(fail_test_path).endswith(pdb_tool.current_frame_file)
+
+
+def test_pdb_crashing(tmp_path, setup_test_repo):
+    "```rewrite 1:2 <c>  asd\n asd</c>```"
```
xingdi-eric-yuan: what's this?

froggy/tools/pdb.py
matheper: Yes, closing session is enough

froggy/tools/pdb.py
sordonia: ok

PR #73, opened 2025-02-21
Better handling of shell session

MarcCote:


froggy/agents/llm_api.py
```diff
@@ -329,7 +329,7 @@ def __call__(self, messages, info, *args, **kwargs) -> LLMResponse:
                 available_commands, ignore_case=True, sentence=True
             )
             action = prompt(
-                "> ",
+                "\n> ",
```
MarcCote, line 332: To avoid conflict with tqdm

froggy/pond/terminal.py
```diff
@@ -14,46 +13,49 @@
...
+        self.default_read_until = self.env_vars["PS1"]  # Disable echo
+
         atexit.register(self.close)
 
     @property
     def is_running(self):
         return self.process is not None and self.process.poll() is None
 
-    def start(self, command=None):
+    def start(self, program=None):  #:str = "/bin/bash --noprofile --norc"):
```
MarcCote: failed attempt at starting a bash within a bash via -c option. Doing that would allow to put `stty -echo` in setup_commands.

froggy/pond/terminal.py
```diff
@@ -358,7 +363,7 @@ def default_shell_command(self) -> list[str]:
         user_map = self.user_map()
         if user_map:
             user_map = f"--user {user_map}"
-        entrypoint = f"docker exec -i {user_map} {self.container.name} /bin/bash --noprofile --norc"
+        entrypoint = f"docker exec -t -i {user_map} {self.container.name} /bin/bash --noprofile --norc"
```
MarcCote, line 373: The -t is needed to see PS1 otherwise there's no prompt -> meaning we can't read_until the PS1

froggy/pond/utils.py
```diff
@@ -155,25 +153,8 @@ def is_subdirectory(path, directory):
...
-
-    def signal_handler(signum, frame):
-        raise TimeoutException("Timed out!")
-
-    signal.signal(signal.SIGALRM, signal_handler)
-    signal.alarm(seconds)
-    try:
-        yield
-    finally:
-        signal.alarm(0)
```
MarcCote, line 176: This was never used. Delete!

PR #81, opened 2025-02-25
Max score

matheper:
- Split eval, score, max_score, done
- Added method to reset env variables

### DIFF
--- a/froggy/pond/envs/env.py
+++ b/froggy/pond/envs/env.py
@@ -173,7 +173,7 @@ class RepoEnv(TooledEnv):
         self.current_file_content = None
         self.current_breakpoints_state = {}
         self.rewrite_counter = 0
-        self.last_eval_output: EvalOutput = None
+        self.last_eval: EvalOutput = None
         self.score = 0
         self.done = False
         # clear all observations and event queue (queue should be empty already)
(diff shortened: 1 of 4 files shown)


froggy/pond/envs/env.py
```diff
@@ -304,23 +307,44 @@ def reset(
...
+
+    def calculate_done(self, eval_output: EvalOutput) -> bool:
+        """Determine if the task is done.
+        Override in subclasses for different behavior."""
+        return self.score == self.max_score
+
+    def cleanup_eval_output(self, eval_output: EvalOutput) -> str:
+        """Cleanup the eval output before setting it as the last_eval_output.
+        Override in subclasses for different behavior."""
+        return eval_output.output
```
MarcCote: That seems very underspecified. I'm not sure if that makes sense to have this in the parent class.

MarcCote: It might be fine if we provided details of what cleanup we expect the user to do when they create a new environment.

froggy/pond/tools/pdb.py
```diff
@@ -147,7 +147,7 @@ def use(self, tool_args) -> Observation:
                     + 1
                 )
                 # TODO: I think this is a shortcut to use the output of pdb instead of eval. Can we remove this?
-                # self.environment.last_eval_obs = output[:end_index]
+                # self.environment.last_eval_output = output[:end_index]
```
MarcCote: Let's remove those

froggy/pond/envs/env.py
matheper: It doesn't do anything here, but I want to implement it in the parent class so we can assume the method exists when we need a clean output, for example in `last_eval_output_text`

froggy/pond/tools/pdb.py
matheper: Is there anything else we could remove from this block of code?

froggy/pond/envs/env.py
matheper: Done. Now eval returns a clean output if required.

PR #97, opened 2025-03-21
LLM API refac

matheper:
This PR do the following:
- Make LLM an abstract class and split each model in it's own class (Azure, OpenAI, Anthropic)
- Move TokenCounter / Tokenizer inside the LLM class
- Add some additional tests
- Some code moving around like the retry_on_rate_limit moved outside LLM

### DIFF
--- a/tests/agents/test_llm_api.py
+++ b/tests/agents/test_llm_api.py
@@ -18,25 +18,6 @@ from debug_gym.agents.llm_api import (
 )
 
 
-@pytest.fixture
-def openai_llm(logger_mock, llm_cfg_mock):
-    # Create an instance of AsyncOpenAILLM with a mock configuration
-    model_name = "test_model"
-    _async_llm = OpenAILLM(model_name, logger=logger_mock)
-    return _async_llm
-
-
-def test_is_rate_limit_error(openai_llm):
-    mock_response = MagicMock()
-    mock_response.request = "example"
-    mock_response.body = {"error": "Rate limit exceeded"}
-    # Instantiate the RateLimitError with the mock response
-    exception = RateLimitError(
-        "Rate limit exceeded", response=mock_response, body=mock_response.body
-    )
-    assert openai_llm.is_rate_limit_error(exception) == True
-
-
 @patch("openai.resources.chat.completions.Completions.create")
 @patch.object(
     LLMConfigRegistry,
@@ -88,11 +69,16 @@ def llm_cfg_mock(tmp_path, monkeypatch):
             }
         )
     )
-    monkeypatch.setenv("LLM_CONFIG_FILE", str(config_file))
     return config_file
 
 
 def test_load_llm_config(llm_cfg_mock):
+    config = LLMConfigRegistry.from_file(config_file_path=str(llm_cfg_mock))
+    assert "test_model" in config
+
+
+def test_load_llm_config_from_env_var(llm_cfg_mock, monkeypatch):
+    monkeypatch.setenv("LLM_CONFIG_FILE_PATH", str(llm_cfg_mock))
     config = LLMConfigRegistry.from_file()
     assert "test_model" in config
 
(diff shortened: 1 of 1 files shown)


debug_gym/agents/llm_api.py
```diff
@@ -61,17 +65,51 @@ def print_messages(messages: list[dict], logger: DebugGymLogger):
...
+        retry=(
+            retry_if_not_exception_type(KeyboardInterrupt)
+            & retry_if_exception(is_rate_limit_error_func)
+        ),
+        wait=wait_random_exponential(multiplier=multiplier, max=max_wait),
+        stop=stop_after_attempt(max_attempts),
+    )
+    return retry_decorator(func)
+
+
```
MarcCote: Love it

debug_gym/agents/llm_api.py
```diff
@@ -61,17 +65,51 @@ def print_messages(messages: list[dict], logger: DebugGymLogger):
...
+    current_role = None
+    for message in messages:
+        if current_role == message["role"]:
+            to_merge.append(message)
+        else:
+            merge()  # merge the previous messages
+            current_role = message["role"]
+            to_merge = [message]
+    merge()  # merge the last messages
     return messages_out
```
MarcCote: Should we move this at the same place as trim_messages (i.e., https://github.com/microsoft/debug-gym/tree/main/debug_gym/agents/utils.py)?

debug_gym/agents/llm_api.py
```diff
@@ -172,158 +156,155 @@ def __init__(self, model_name: str, logger: DebugGymLogger | None = None):
...
+        """Abstract method to tokenize a text."""
+        pass
+
+    def token_counter(self, text: str) -> int:
+        """Count the number of tokens in a text."""
+        return self.count_tokens(text)
+
+    def count_tokens(self, text: str) -> int:
+        """Count the number of tokens in a text."""
+        return len(self.tokenize(text))
```
MarcCote: Any reason we have both? Just to avoid changing other files?

debug_gym/agents/llm_api.py
```diff
@@ -337,7 +318,9 @@ def query_anthropic_model(self, messages, **kwargs):
             kwargs["temperature"] = 1.0
 
             response = (
-                self.call_with_retry(self.client.messages.create)(
+                retry_on_rate_limit(
+                    self.client.messages.create, self.is_rate_limit_error
+                )(
                     model=self.config["model"],
                     thinking={"type": "enabled", "budget_tokens": 16000},
```
MarcCote: Let's put `budget_tokens` as a parameter instead of hardcoded. @xingdi-eric-yuan thoughts?

debug_gym/agents/llm_api.py
```diff
@@ -350,7 +333,9 @@ def query_anthropic_model(self, messages, **kwargs):
         else:
             kwargs["max_tokens"] = 8192
```
MarcCote: Where does this default value comes from?

debug_gym/agents/llm_api.py
```diff
@@ -370,125 +355,171 @@ def query_anthropic_model(self, messages, **kwargs):
...
+                    self.logger.debug(
+                        f"Error calling {self.model_name}: {exception_full_name!r} {exception.message}"
+                    )
+                    return True
+                else:
+                    self.logger.warning(
+                        f"Error calling {self.model_name}: {exception_full_name!r} {exception.message}"
+                    )
+                    return False
+            case _:
```
MarcCote: I don't know what's the best way to handle this, those special errors come from using Trapi. This wouldn't be relevant to anyone outside MSR folks :/.

debug_gym/agents/llm_api.py
matheper: No, just to open the PR and gather some feedback. I'm removing `token_counter` next.

debug_gym/agents/llm_api.py
matheper: @xingdi-eric-yuan might know

debug_gym/agents/llm_api.py
matheper: Sounds good. I'm going to move `print_messages` as well.

debug_gym/agents/llm_api.py
matheper: I'm wondering if all these things could go to `llm.cfg` or `config.yaml`.
Then llm_api just pass the configs as kwargs.

By `all things` I mean things like:
```
if "thinking" in self.config.get("tags", []):
    kwargs["max_tokens"] = 20000
    kwargs["temperature"] = 1.0
    ...
    thinking={"type": "enabled", "budget_tokens": 16000},
...
else:
    kwargs["max_tokens"] = 8192
    ...
```

debug_gym/agents/llm_api.py
matheper: Probably fine if we're not skipping retries for legit cases. Do you know what cases fall into the no-retry bucket (that else statement)?

matheper: Maybe better moving this logic to AzureOpenAILLM

debug_gym/agents/llm_api.py
xingdi-eric-yuan: This number was from @Kim-Minseon 's branch.

debug_gym/agents/llm_api.py
matheper: Addressed in PR #99.

debug_gym/agents/llm_api.py
matheper: PR #99 keeps the same values, but as a config in `llm.cfg`.

debug_gym/agents/llm_api.py
matheper: There's a small refact for clarity, but I'll keep it where it is.

debug_gym/agents/llm_api.py
```diff
@@ -276,6 +237,7 @@ def is_rate_limit_error(self, exception) -> bool:
         rate_limit_errors = [
             "anthropic.RateLimitError",
             "anthropic.OverloadedError",
+            "anthropic._exceptions.OverloadedError",
```
MarcCote, line 370: What is this "OverloadedError" ?

matheper: @xingdi-eric-yuan I integrated this from #96? Do you what that is?

matheper: When Anthropic server is busy with too many requests. `OverloadedError` is not in the `_exceptions` `__all__` list, that's why we access it through `_exceptions`. I'll keep both for now.

PR #99, opened 2025-03-24
Llm config

matheper:
Data classes for LLM Configs
Get generate_kwargs from config

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,14 +18,16 @@ To install the development dependencies:
 
 First, make a copy of the template,
 
-    cp llm.template.yaml llm.yaml
+    cp llm.template.yaml ~/.config/debug_gym/llm.yaml
 
-Then, edit llm.yaml with your endpoint and credentials. You can choose one of these authentication methods:
+Then, edit this file with your endpoint and credentials. You can choose one of these authentication methods:
 - For authenticating with an API key, provide `api_key`.
-- For `az login` or Managed Identity authentication, remove `api_key` and include `scope` instead.
+- For `az login` or Managed Identity authentication on Azure, remove `api_key` and include `scope` instead.
 
 > [!WARNING]
 > When using open-sourced LLMs, e.g., via vLLM, you need to correctly setup `HF_TOKEN` required by the tokenizer.
+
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
MarcCote, line 1: Why are we pushing this file?

debug_gym/agents/llm_api.py
```diff
@@ -75,6 +66,95 @@ def retry_on_rate_limit(
...
+        # Convert each model configuration to LLMConfig objects
+        for model_name, model_config in configs.items():
+            registry.register(model_name, model_config)
+        return registry
+
+    @classmethod
+    def from_file(cls, config_file_path: str | None = None) -> "LLMConfigRegistry":
+        """Load the LLM configuration from a JSON file"""
+        if config_file_path is None:
+            config_file_path = os.environ.get("LLM_CONFIG_FILE", "llm.cfg")
```
MarcCote: At some point, we should store it in the ~/.config/debug-gym/llm.cfg

debug_gym/agents/llm_api.py
```diff
@@ -75,6 +66,95 @@ def retry_on_rate_limit(
...
+        try:
+            with open(config_file_path) as f:
+                raw_llm_config = json.load(f)
```
MarcCote: Can we switch to .yaml file instead of .json?

.vscode/settings.json
matheper: No reason, let me remove it.

debug_gym/agents/llm_api.py
matheper: Done

debug_gym/agents/llm_api.py
matheper: Done!

README.md
```diff
@@ -14,19 +14,21 @@ To install the development dependencies:
 
     pip install -e '.[dev]'
 
-**Set your API information in llm.cfg**
+**Set your API information in llm.yaml**
 
 First, make a copy of the template,
 
-    cp llm.cfg.template llm.cfg
+    cp llm.template.yaml ~/.config/debug_gym/llm.yaml
```
MarcCote, line 21: Need to create the debug_gym folder first, I think.

MarcCote: If we really want it to be clean, we could create that file (initialized to the template content) on pip installing the library.

matheper: Makes sense, let's initialize the template in a separate PR.

PR #107, opened 2025-03-28
HITL agent

chisingh:


debug_gym/gym/envs/env.py
```diff
@@ -251,6 +251,10 @@ def set_entrypoints(self, entrypoint, debug_entrypoint):
 
     @staticmethod
     def _prepare_entrypoint(entrypoint):
+        # Remove PYTHONPATH prefix if present
+        if entrypoint.startswith("PYTHONPATH=$PYTHONPATH:$PWD "):
+            entrypoint = entrypoint[len("PYTHONPATH=$PYTHONPATH:$PWD ") :]
+
```
MarcCote: Why is that?

chisingh: It was causing the `if entrypoint_list[0] != "python":` check to fail. The cloned environment later gets this path updated in the `set_entrypoints()`.

debug_gym/gym/envs/env.py
```diff
@@ -547,6 +547,23 @@ def step(self, action: ToolCall) -> EnvInfo:
...
+        # Create a new instance of RepoEnv
+        new_env = RepoEnv(
+            path=self.path,
+            entrypoint=self.entrypoint,
+            debug_entrypoint=self.debug_entrypoint,
+            max_score=self.max_score,
+            readonly_patterns=None,
+            run_timeout=self.run_timeout,
+            dir_tree_depth=self.dir_tree_depth,
+            terminal=Terminal(),
```
MarcCote: We have two type of terminal: Terminal and DockerTerminal. The clone should be using the right one.

debug_gym/gym/envs/env.py
```diff
@@ -547,6 +547,23 @@ def step(self, action: ToolCall) -> EnvInfo:
 
         return self.infos
 
+    def clone(self):
```
MarcCote: Should we be subclassing the clone method in all type of envs? Some environments (e.g., SWE-Smith) require custom setup.

MarcCote: Ideally, we would have a generic way to create a copy of the environment. Maybe by reusing the the logic in run.py

debug_gym/gym/envs/env.py
```diff
@@ -547,6 +547,23 @@ def step(self, action: ToolCall) -> EnvInfo:
...
+            debug_entrypoint=self.debug_entrypoint,
+            max_score=self.max_score,
+            readonly_patterns=None,
+            run_timeout=self.run_timeout,
+            dir_tree_depth=self.dir_tree_depth,
+            terminal=Terminal(),
+            logger=self.logger,
+        )
+        for tool in self.tools:
+            new_env.add_tool(tool)
```
MarcCote: I'm worry that that a same tool would be use for multiple environment. Are we sure that's safe?

MarcCote (changes requested):
When I tried it, I notice the RewriteOnly agent didn't have the proper context.

debug_gym/llms/human.py
```diff
@@ -470,6 +471,12 @@ def __init__(
         if prompt_toolkit_available:
             self._history = InMemoryHistory()
 
+        # Warn if self.logger.level is not set at least to INFO, as this is a human interface.
+        if self.logger.level > logging.INFO:
```
matheper, line 476: Is this required for the human mode to work / users to see the outputs? If so, should we raise an exception instead of just logging a warning?

MarcCote: Good point. Yes, we should raise an exception. It doesn't make sense to run human mode without seeing the INFO logs.

PR #141, opened 2025-06-02
WIP Trajectory Filtering

xingdi-eric-yuan:


MarcCote:
Is that ready to review?

analysis/trajectory_filtering/filter.py
```diff
@@ -0,0 +1,152 @@
+import json
+import os
+
+from criteria import *
+from tqdm import tqdm
+
+exp_path = "../../exps/"
```
matheper: Can you get these two values from the command line?

analysis/tool_use/episode_length_tokens.py
```diff
@@ -0,0 +1,281 @@
...
+    plt.legend.loc = "upper left"
+
+    plt.grid(True, alpha=0.3)
+    plt.tight_layout()
+    plt.show()
+
+
+# Example usage:
+model_paths = [
+    "../../exps/may22/rewrite_4o",
```
matheper: Would it be possible to iterate over the content of this directory instead of listing the values here? Also, could you get the directory from an arg in the command line? The same comment applies to the other files.

CHANGELOG.md
```diff
@@ -1,25 +1,25 @@
 # Changelog
 
+### 2025-06-11
```
matheper, line 8: Nice, thanks for keeping this up to date!

analysis/tool_use/episode_length_tokens.py
xingdi-eric-yuan: I was a bit hesitant to do this because there're cases when I don't want to show all experiments in the path. 
update: I think I found a way, working on it.

xingdi-eric-yuan: done

matheper: Nice! Was the way adding `--exp-uuid-list`?

xingdi-eric-yuan: yeah `--exp-path` for the log folder, `--exp-uuid-list` for the set of sub-folders to include.

matheper: Do we need to merge the changes to `tech_report_v1` folder?
I would either not change them or get them from args like you did here.

xingdi-eric-yuan: For those I'd keep them there (or maybe even remove from main because they are outdated (and the runnable versions are in v1.0.0). What do you think?

matheper (approved):

PR #143, opened 2025-06-03
Use better command completion for Human Mode

MarcCote:
This pull request improves the user experience when using the Human mode, e.g. `-p base.llm_name=human`.

https://github.com/user-attachments/assets/332bbca2-7158-42d1-a234-3f02e2f84877


xingdi-eric-yuan (approved):
vibe master

scripts/config_mini_nightmare.yaml
```diff
@@ -18,7 +18,7 @@ base:
         # session_commands define commands that are always executed before starting a shell session or running a single command in the terminal.
         # session_commands:["conda activate aider"],
         # setup_commands define commands that are executed only once when the terminal is created. This is only supported for Docker terminal.
-        setup_commands: ["pip install pytest"],
+        setup_commands: ["pip install pytest pandas"],
```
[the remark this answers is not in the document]
MarcCote: It's not for human mode per se. It is for the mini_nightmare.

tests/llms/test_human.py
```diff
@@ -1,4 +1,4 @@
-import json
+# Disclaimer: Most tests in this file are generated by the VSCode Agent with Claude Sonnet 4.
```
matheper, line 1: Do they all make sense? Sometimes these generated tests are not very useful, as they just mock and assert things without testing any real functionality

MarcCote: I eyeballed a couple and they made sense. Let me have a more thorough look.

matheper:
That's absolutely minor, but the error message is wrong here:

Screenshot 2025-06-04 at 2 17 21 PM

tests/llms/test_human.py
MarcCote: All good.

PR #151, opened 2025-06-11
replace unescape by filtering non-utf8 chars in system prompts

xingdi-eric-yuan:
Now the system prompt content should be able to be parsed by `json.loads()` into a valid JSON object.


matheper:
Can you remove the unescape function and its test if not used anywhere else and add a small test for this new filter_non_utf8.

xingdi-eric-yuan:
> Can you remove the unescape function and its test if not used anywhere else and add a small test for this new filter_non_utf8.

The unescape function is still used [here](https://github.com/microsoft/debug-gym/blob/remove_unescape/debug_gym/gym/utils.py#L11), when we pp the code. 
I'll create test for the new funcion.

matheper (approved):

PR #154, opened 2025-06-11
Only load image for instance_id we want to tests

MarcCote:


.github/workflows/tests.yml
```diff
@@ -30,11 +30,18 @@ jobs:
...
+            debug_gym/gym/envs/swe_*.py
+            tests/gym/envs/test_swe_*.py
       - name: Test with pytest - PR
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && steps.changed-files-specific.outputs.any_changed != 'true'
         run: |
           DEBUG_GYM_DEBUG=1 pytest -vv -n 16 -k "not test_swe_bench and not test_swe_smith" --cov=debug_gym --cov-report=term-missing --cov-fail-under=80 --timeout=300
       - name: Test with pytest
-        if: github.event_name != 'pull_request'
+        if: github.event_name != 'pull_request' || steps.changed-files-specific.outputs.any_changed == 'true'
```
MarcCote: @matheper what do you think of this?

matheper: Looks good, do we need the second condition `|| steps.changed-files-specific.outputs.any_changed == 'true'`?

.github/workflows/tests.yml
```diff
@@ -30,11 +30,18 @@ jobs:
...
+      - name: Get changed files related to SWE-Bench or SWE-Smith
+        id: changed-files-specific
+        uses: tj-actions/changed-files@v46.0.5
+        with:
+          files: |
+            debug_gym/gym/envs/swe_*.py
+            tests/gym/envs/test_swe_*.py
       - name: Test with pytest - PR
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && steps.changed-files-specific.outputs.any_changed != 'true'
```
matheper, line 41: Probably requires increasing the timeout, which is fine. The idea was just to prevent any potential infinite loop or anything like that. Let's set both at 600 sec.

MarcCote: What about this?

matheper: Nice! That's great!

matheper (approved):

PR #159, opened 2025-06-23
Resolve path mismatch issue raised on macOS

dkokkotas:
On macOS, all copied files under `/var` are actually symlinked to `/private/var`.
Currently, `resolve_path()` calls `.resolve()`, and that being the case, a tool-supplied relative filepath (e.g., `test.py`) is converted & expanded to `/private/var/folders/.../RepoEnv-XXXX/test.py`, while the working directory of the corresponding `RepoEnv` instance remains `/var/folders/.../RepoEnv-XXXX`. 
As expected, a consequent `FileNotFoundError` is raised, even though the file is present.

path-mismatch

By letting the working_dir to directly use `.resolve()`, we ensure that it is fully resolved to its canonical path, avoiding symlink-related mismatches.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,11 +12,12 @@ on:
 jobs:
   tests:
     name: Test
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         language: [ "python" ]
+        os: [ubuntu-latest, macos-latest]
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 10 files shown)


MarcCote (approved):
Works on Linux. Good for me for now.

matheper (approved):

matheper:
I've pushed some changes to #164, could you please merge it once again. I believe all tests will pass once those changes are merged. `debug-gym` still isn't fully compatible with macOS, though. The interactive terminal session used by the `pdb` tool relies on Python's `pty`, which is not supported when using Docker on non-Linux platforms.

matheper:
@dkokkotas let's skip the macOS pipeline for now. There are some differences in how the shell outputs, which are unrelated to your contribution, sorry about that. Could you please remove `macos-latest` from the matrix.os list in `.github/workflows/tests.yml`?

PR #162, opened 2025-06-23
Context chang for sft

xingdi-eric-yuan:


debug_gym/gym/envs/env.py
```diff
@@ -160,6 +160,7 @@ def __init__(
         auto_list: bool = True,
         terminal: Terminal | None = None,
         logger: DebugGymLogger | None = None,
+        **kwargs,
```
matheper, line 163: Just for my own understanding, why are the kwargs needed here?

debug_gym/gym/tools/rewrite.py
```diff
@@ -16,7 +16,7 @@ class RewriteTool(EnvironmentTool):
         """rewrite(path='code/utils.py', start=4, end=6, new_code="        print('buongiorno')") will replace the chunk of code between line number 4 and 6 in the specified file 'code/utils.py' by the single line provided, with the indent ahead (in this case, 8 spaces).""",
     ]
     description = (
-        "Rewrite the content of the specified file path, between lines [start, end], with the new code. Line numbers are 1-based. When start and end are None, it's assumed to rewrite the whole file. When start is provided and end is None, it's assumed to rewrite a single line (start). The new code should be valid python code include proper indentation (can be determined from context)."
+        "Rewrite the content of the specified file path, between lines [start, end], with the new code. Line numbers are 1-based. When start is provided and end is None, it's assumed to rewrite a single line (start). When both start and end are None, it's assumed to rewrite the whole file, this is not recommended because most of the time the expected edit is local. The new code should be valid python code include proper indentation (can be determined from context)."
```
matheper, line 19: Was the description for when `end` is not provided incorrect? Good catch!

debug_gym/gym/envs/env.py
xingdi-eric-yuan: Becasue I have new kwargs like `show_current_breakpoints` and `show_directory_tree` in the `env_kwargs` in config. But these args are not used in the env, only at where we create prompt.

debug_gym/gym/tools/rewrite.py
xingdi-eric-yuan: No it's just to discourage rewriting the entire file.

matheper: The previous description was "rewrite the whole file", but it only rewrites a single line.

xingdi-eric-yuan: ah I swapped the two cases (both None vs end is None), to make the description more natural.

PR #167, opened 2025-06-26
Improve filter

xingdi-eric-yuan:


debug_gym/llms/openai.py
```diff
@@ -214,10 +214,15 @@ def generate(self, messages, tools, **kwargs) -> LLMResponse:
...
         except openai.BadRequestError as e:
             # Handle specific error for context length exceeded, otherwise just propagate the error
-            if e.code == "context_length_exceeded":
+            if e.code in [
+                "context_length_exceeded",
+                "model_max_prompt_tokens_exceeded",
+                "string_above_max_length",
+            ]:
+                raise ContextLengthExceededError
+            if "maximum context length" in e.message:
```
matheper: Just curious, what is the e.code in this case?

debug_gym/gym/envs/env.py
```diff
@@ -109,6 +109,9 @@ def get_triggered_tools(self, action: ToolCall):
         except Exception as e:
             # parse error
             return str(e), None
+        if tool_name == "empty_tool_response":
+            # the agent did not generate any tool call
+            return "No tool call was generated by the agent.", None
```
matheper, line 114: Is "agent" mentioned anywhere else?

debug_gym/llms/base.py
```diff
@@ -347,24 +347,58 @@ def generate_with_drop_message_and_retry(messages, tools, **kwargs):
...
-                llm_response = generate_with_drop_message_and_retry(
-                    messages, tools, **kwargs
-                )
-                self.logger.info(
-                    f"Prompt truncated to {llm_response.token_usage.prompt:,} tokens."
-                )
 
-            print_messages(messages, self.logger)
-            return llm_response
+            max_retries = 3  # Prevent infinite recursion
```
matheper: Does this mean at most three messages are dropped? Should this value be higher?

xingdi-eric-yuan: No ideally a single iteration could be enough, because it will 1) drop multiple message turns so the output could fit into the context length; or 2) trigger some assertion, e.g., the system message itself exceeds the context length. 

So maybe we can set max_retries to be 1 even...

The inf loop before was mainly because the mismatch between tokenizers, especially with Anthropic models where we don't have a tokenizer available, the best thing we could do was to use a tiktoken to estimate. This caused the case where after truncation, the message still being rejected from LLM call.

debug_gym/gym/envs/env.py
xingdi-eric-yuan: Yes in the default sys prompt we say `You are a debugging agent specialized in fixing Python programs...`

debug_gym/llms/openai.py
xingdi-eric-yuan: added the code name
```
Response JSON: {
  "error": {
    "message": "This model's maximum context length is 128000 tokens. However, your messages resulted in 164408 tokens. Please reduce the length of the messages.",
    "code": "invalid_request_body"
  }
}
```

matheper: Thanks for adding the e.code to the if condition! 😄

debug_gym/llms/base.py
matheper: Makes sense. We can remove the for loop in this case.

xingdi-eric-yuan: I kept the loop of 1 to avoid repeating code (the llm.generate) when retry

matheper (approved):

PR #169, opened 2025-07-02
Agent as tool

sordonia:
- adds agent as tool capability

### DIFF
--- a/debug_gym/gym/tools/agent.py
+++ b/debug_gym/gym/tools/agent.py
@@ -1,0 +1,83 @@
+import logging
+import os
+from os.path import join as pjoin
+
+from debug_gym.agents.history_tracker import build_history_prompt
+from debug_gym.gym.entities import Observation
+from debug_gym.gym.tools.tool import EnvironmentTool
+from debug_gym.gym.tools.toolbox import Toolbox
+from debug_gym.gym.utils import is_subdirectory, show_line_number
+from debug_gym.llms.base import LLM
+
+
+@Toolbox.register()
+class AgentTool(EnvironmentTool):
+    name: str = "agent"
+    examples = [
+        """agent(query="") Executes query to send to this specialized agent.\n"""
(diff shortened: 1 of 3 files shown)


xingdi-eric-yuan:
I feel the `agentic_tool` can be something more generic, and this history summarization tool being an instantiation of such class?

debug_gym/gym/tools/agent.py
```diff
@@ -0,0 +1,81 @@
...
+from debug_gym.gym.tools.toolbox import Toolbox
+from debug_gym.gym.utils import is_subdirectory, show_line_number
+from debug_gym.llms.base import LLM
+
+
+@Toolbox.register()
+class AgentTool(EnvironmentTool):
+    name: str = "agent"
+    examples = [
+        """agent(query="") Executes query to send to this specialized agent.\n"""
```
MarcCote: What do you mean by "Executes query".
```suggestion
        """agent(query="") Send a query to this specialized agent.\n"""
```

MarcCote (changes requested):
Left a comment to be addressed before merging.

debug_gym/gym/tools/agent.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+    arguments = {
+        "query": {
+            "type": ["string"],
+            "description": "The request to the agent.",
+        },
+    }
+
+    def __init__(
+        self, history, llm_name=None, llm_config=None, llm_config_file_path=None
```
MarcCote, line 29: What is this `history`? A reference to an agent's `history`?

sordonia: it's the HistoryTracker of the caller agent, the agentic tool might want to peek at the interactions of the caller agent

MarcCote: But do you have an example of how it is use?

PR #170, opened 2025-07-02
Rich logger

matheper:


xingdi-eric-yuan:
Do we want to add the step/episode timeout in this PR or in another?

matheper:
We can work on a separate PR for the step timeout and keyboard interruption.

debug_gym/logger.py
```diff
@@ -39,17 +338,37 @@ def __init__(
...
-        console.setFormatter(formatter)
-        console.setLevel(level)
-        self.addHandler(console)
+        # Prevent the log messages from being propagated to the root logger
+        self.propagate = False
+
+        self.level = level
+        self.setLevel(self.level)
+        self.log_file = None  # File handler for logging to a file
+        self._live = None  # rich live context manager for updating the UI
```
MarcCote, line 354:
```suggestion
        self._live = None  # rich live context manager for updating the UI
        self._log_listener_thread = None  # Thread for listening to log messages
```

MarcCote (approved):

debug_gym/logger.py
matheper: Good catch, done!

PR #171, opened 2025-07-03
Fix init obs

xingdi-eric-yuan:
In previous versions of the code, the `build_history_prompt` method was filtering out actions with a `None` value. This will cause the agent to miss the initial observation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -184,3 +184,4 @@ vscode/out
 vscode/node_modules
 vscode/package-lock.json
 .vscode
+debug_gym/llms/claude.py
--- a/debug_gym/agents/base_agent.py
+++ b/debug_gym/agents/base_agent.py
@@ -58,7 +58,7 @@ class BaseAgent:
 
     def build_history_prompt(self):
         messages = build_history_prompt(
-            self.history.filter_out(actions=[None]),
+            self.history,
             self.llm,
             self.config["reset_prompt_history_after_rewrite"],
         )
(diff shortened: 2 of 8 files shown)


debug_gym/agents/base_agent.py
```diff
@@ -58,7 +58,7 @@ def set_seed(self, seed):
 
     def build_history_prompt(self):
         messages = build_history_prompt(
-            self.history.filter_out(actions=[None]),
+            self.history,
```
matheper, line 61: I think we can remove the `HistoryTracker.filter_out` method as well.

.gitignore
```diff
@@ -184,3 +184,4 @@ vscode/out
 vscode/node_modules
 vscode/package-lock.json
 .vscode
+debug_gym/llms/claude.py
```
matheper, line 187: do we need this?

debug_gym/llms/anthropic.py
```diff
@@ -111,35 +111,54 @@ def parse_tool_call_response(self, response) -> ToolCall:
...
-                        "tool_use_id": history_info.action.id,  # 'toolu_01SdR84CsnTKRpdH4zwFjvGj'
-                        "content": filter_non_utf8(
-                            f"{history_info.step_observation.observation}"
-                        ),  # 'Viewing `hangman_test.py`. The file is read-only, it is not editable.'
-                    }
-                ],
-            },
-        ]
+        _messages = []
+        if isinstance(response, list) and len(response) > 0:
```
matheper, line 117: It seems we never wrote tests for this method. Could you write tests for the different cases showing the possible inputs and expected messages?

matheper (approved):
LGTM!

PR #176, opened 2025-07-09
Add get_problem_ids (formerly get_dataset_split) to all benchmark env

MarcCote:
This pull request introduces a new `get_problem_ids` method across multiple environments to unify the way problem IDs are retrieved and validated. It also updates the relevant code and adds comprehensive tests to ensure the method's correctness and robustness.

### New method implementation:

* Added `get_problem_ids` method to the following files to standardize problem ID retrieval:
  - `debug_gym/gym/envs/aider.py`
  - `debug_gym/gym/envs/swe_bench.py`
  - `debug_gym/gym/envs/swe_smith.py`, replacing the previous `get_dataset_split` method

### Code updates for integration:

* Updated `scripts/run.py` to use the new `get_problem_ids` method instead of `get_dataset_split`. This ensures consistency across environments.

### Test coverage improvements:

* Added unit tests for `get_problem_ids` in the following test files:
  - `tests/gym/envs/test_aider.py`
  - `tests/gym/envs/test_swe_bench.py`
  - `tests/gym/envs/test_swe_smith.py`

### DIFF
--- a/scripts/run.py
+++ b/scripts/run.py
@@ -103,10 +103,7 @@ def main():
     problems = config.get("problems", ["custom"])
     if isinstance(problems, str) and "benchmark" in config:
         env = create_env(config, logger=logger)
-        if problems == "all":
-            problems = sorted(env.dataset.keys())  # all tasks
-        else:
-            problems = env.get_dataset_split(problems)
+        problems = env.get_problem_ids(split_or_problem_id=problems)
 
     if args.list:
         print(f"\n# Available problems in {config.get('benchmark', 'config')}:")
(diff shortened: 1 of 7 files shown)


scripts/run.py
```diff
@@ -103,10 +103,7 @@ def main():
     problems = config.get("problems", ["custom"])
     if isinstance(problems, str) and "benchmark" in config:
```
matheper, line 104: Can you remind me why we check for "benchmark" here?

matheper: Can we move everything related to the problems selection into `get_problem_ids`?

Maybe this is too much for this PR, but would it be possible to split the code that load the `env.dataset` from the rest of `env.load_dataset`, so we could expose `get_problem_ids` as a class method?

MarcCote: Yeah, I think we need to revisit the whole load_dataset thing as we discussed previously.

matheper: Sounds good!

PR #190, opened 2025-07-14
Add logfile to progress tracker. Stop pending tasks' spinner.

MarcCote:


debug_gym/logger.py
```diff
@@ -113,12 +114,18 @@ def __init__(self, spinner_name: str = "dots", speed: float = 1.0):
 
     def render(self, task):
         status = task.fields.get("status")
+        if status is None:
+            return Text(
+                TaskProgress.marker("pending"),
+                style=TaskProgress.color("pending"),
+            )
+
```
MarcCote: @matheper do you know why the status of pending task is `None`?

debug_gym/logger.py
```diff
@@ -376,7 +387,7 @@ def status_report(self):
     def _status_listener(self):
         self.logger.debug("Starting status listener thread...")
         last_refresh_time = 0
-        refresh_interval = 1  # Minimum time between UI refreshes (in seconds)
+        refresh_interval = 0  # 1  # Minimum time between UI refreshes (in seconds)
```
MarcCote: This prevented the overall progress tracker to show the accurate number of running tasks (because all the workers start at the same time).

debug_gym/logger.py
```diff
@@ -443,7 +454,7 @@ def __init__(
             self.addHandler(fh)
 
     def _initialize_main_logger(self):
-        self._live = Live(transient=True)
+        self._live = Live(transient=True, refresh_per_second=2)
```
MarcCote, line 447: This was needed to make the ctrl+click on log path in vscode terminal work better. With 2 refreshes per second, you can still get an okay spinner animation.

scripts/run.py
```diff
@@ -33,6 +33,16 @@ def run_agent(args, problem, config):
...
+
+        task_logger.report_progress(
+            problem_id=problem,
+            step=0,
+            total_steps=1,
+            score=0,
+            max_score=100,
+            status="running",
+        )
+
```
MarcCote, line 67: When the worker start, signal the progress tracker that it is now running and show the path to the log file.

debug_gym/logger.py
matheper: Sorry, this code is a bit confusing because it uses task to refer to both rich `Task` the `TaskProgress` dataclass. This one is a `rich Task`, the status is empty until the logger processes the first `TaskProgress` message at `TaskProgressManager.advance`

matheper: Actually, you can probably set the task status when adding the task and assume rich Tasks aways have a status. Something like:
```
pid = self.progress.add_task(
    ...
    fields={"status": task.status},
)
```

debug_gym/logger.py
matheper: No problem. I'm fine with even slower refresh rate if that helps the ctrl+click.

debug_gym/logger.py
matheper: I'm ok removing all this `last_refresh_time` and `refresh_interval` logic if there's no issue rendering the tasks. I believe I implemented that while the code was not quite right and rendering was taking too long to complete.

debug_gym/logger.py
matheper: You can use `fields` to also set `logfile=""`.

MarcCote: Is that better than setting it up in TaskProgress?

matheper: No, it's better to setting it up in TaskProgress. It'll still require both, but you can safely access `task.fields.status` and `task.fields.logfile` at any time if you set them with some default value at initialization.

debug_gym/logger.py
```diff
@@ -113,12 +114,12 @@ def __init__(self, spinner_name: str = "dots", speed: float = 1.0):
 
     def render(self, task):
         status = task.fields.get("status")
-        if task.finished:
+        if task.finished or status == "pending":
```
matheper: Two questions:
1. Can you invert the logic? `if running`.
2. Do you still need this `get/if status` if you set tasks.fields.status at initialization?

MarcCote: 1. I can
2. I think we can drop the .get but I want the pending to be a fixed text. Maybe there's a better way to stop the spinning animation?

matheper (approved):
Thanks for fixing the report progress too!

PR #192, opened 2025-07-16
PyPI release action

matheper:


.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+          pip install -e '.[dev]'
+
+      - name: Save version and commit hash
+        run: |
+          VERSION="$(git describe --tags --abbrev=0 | sed 's/^v//')"
+          COMMIT="$(git rev-parse --short HEAD)"
+          echo "__version__ = \"${VERSION}\"" > debug_gym/version.py
+          echo "__commit__ = \"${COMMIT}\"" >> debug_gym/version.py
+          echo "__version__ = \"${VERSION}\""
+          echo "__commit__ = \"${COMMIT}\""
```
MarcCote: leftover?

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Build package
+        run: python -m build
+
+      # - name: Publish to PyPI
+      #   uses: pypa/gh-action-pypi-publish@release/v1
```
MarcCote: Not ready for prime time?

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
+name: PyPI Release
+
+# on:
+#   push:
+#     tags:
+#       - 'v*'  # Only publish on version tags
```
MarcCote: smart but anyone with push access can make a release? Is there a way we can constraint on the main branch, since that's protected?

matheper: I believe we can add rules to protect tags, let me check.

.github/workflows/release.yml
matheper: Kind of still debugging the pipeline 😅

matheper: Removing this for now. It will still require bumping the version manually.

.github/workflows/release.yml
matheper: There's a new rule to require admin access to manage tags.

.github/workflows/release.yml
matheper: Ready!

MarcCote (approved):

PR #193, opened 2025-07-17
Dump experiment info

matheper:
Log detailed experiment metadata, including versioning and configuration details, in the `scripts/run.py` file. The changes include adding a new utility function, `dump_experiment_info`, and updating the main script to ensure experiment metadata is saved in a structured format.

**New `dump_experiment_info` function**: Added a utility function to log experiment metadata, including the `debug-gym` version, current datetime, Git commit hash, Git diff, Python version, and experiment configuration. The metadata is saved in a JSONL file within the experiment's output directory. (`[scripts/run.pyR191-R244](diffhunk://#diff-31f00059cd1ed0d328e70d7f85c8c908e9252130223c372c3e24397c3ebe06a0R191-R244)`)

### DIFF
--- a/scripts/run.py
+++ b/scripts/run.py
@@ -1,10 +1,13 @@
+import datetime
 import json
 import os
 import signal
+import subprocess
 import uuid
 from concurrent.futures import ProcessPoolExecutor, as_completed
 from pathlib import Path
 
+from debug_gym import version as dg_version
 from debug_gym.agents.base_agent import AGENT_REGISTRY, create_agent
 from debug_gym.agents.utils import load_config
 from debug_gym.gym.envs import select_env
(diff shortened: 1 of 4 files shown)


scripts/run.py
```diff
@@ -185,12 +188,60 @@ def add_tools(env, config: dict, logger: DebugGymLogger):
...
+        "datetime": datetime.datetime.now().isoformat(),
+        "git_hash": git_hash,
+        "git_diff": git_diff,
+        "config": config,
+        "args": vars(args),
+        "python_version": os.sys.version,
+    }
+
+    file = Path(config["output_path"]) / config["uuid"] / "experiment_info.jsonl"
+    with open(file, "a") as f:
```
MarcCote, line 224: Good call, to have the append here.

scripts/run.py
```diff
@@ -185,12 +188,60 @@ def add_tools(env, config: dict, logger: DebugGymLogger):
...
 
 
+def dump_experiment_info(config: dict, args: dict):
+    """Dump experiment information to a JSONL file.
+    Each line is one experiment run with its metadata."""
+
+    try:  # Get debug-gym version
+        debug_gym_version = dg_version.__version__
+    except Exception:
+        debug_gym_version = ""
```
MarcCote: Can't we assume now, that `__version__` will always exist?

scripts/run.py
```diff
@@ -185,12 +188,60 @@ def add_tools(env, config: dict, logger: DebugGymLogger):
...
+        )
+    except Exception:
+        git_hash = ""
+
+    try:  # Get git diff
+        git_diff = subprocess.check_output(
+            ["git", "diff"], cwd=os.path.dirname(__file__)
+        ).decode()
+    except Exception:
+        git_diff = ""
```
MarcCote, line 211: Just checking. If a user pip install debug-gym, and they are running experiment in their own git codebase base, this will log their commit hash and git diff? Not saying that's a bad thing, just checking if that was intentional. In fact, that might be a good thing.

MarcCote (approved):
Will there be another PR to get track of status of the tasks as well?

scripts/run.py
matheper: Yes, we can. I'll remove it. I let the try except there just because I had no clear preference about what to do with it.

scripts/run.py
matheper: That's correct. I thought about that and reached the same conclusion as you. I don't think it's worth implementing anything to prevent that from happening.

PR #197, opened 2025-07-21
Dump problem status

matheper:
Dump problem status to json and load it back if file exists


debug_gym/logger.py
```diff
@@ -163,6 +196,20 @@ def __init__(self, problems, max_display: int = 10) -> None:
...
+                log_file_path(
+                    progress_update.logdir,
+                    progress_update.problem_id,
+                    relative=True,
+                )
+            )
+            if progress_update.logdir
+            else ""
+        )
+
```
MarcCote: Why is this not a method of TaskProgress?

MarcCote (approved):
LGTM. I'll wait for @xingdi-eric-yuan to have a look and maybe give it a try.

debug_gym/logger.py
matheper: Good point, I'll move it to TaskProgress.

PR #201, opened 2025-07-25
Qwen reasoning content

xingdi-eric-yuan:


debug_gym/llms/openai.py
```diff
@@ -234,9 +234,20 @@ def generate(self, messages, tools, **kwargs) -> LLMResponse:
...
+        # In openai call, the content is in response.choices[0].message.content
+        # In some models hosted on vllm, e.g., qwen-3, there could be content in both (when reasoning is enabled)
+        # response.choices[0].message.content and response.choices[0].message.reasoning_content
+        # https://qwen.readthedocs.io/en/latest/deployment/vllm.html#parsing-thinking-content
+        _content = response.choices[0].message.content or ""
+        if hasattr(response.choices[0].message, "reasoning_content"):
+            reasoning_content = response.choices[0].message.reasoning_content
+            # Only add reasoning content if it's a string (not a MagicMock for tests)
+            if isinstance(reasoning_content, str) and reasoning_content:
+                _content = "\n".join([_content, reasoning_content]).strip()
```
sordonia, line 246: the reasoning content might be the content inside <think> </think> tokens

do we want to consider that as normal content?

PR #203, opened 2025-07-26
Grep Tool

sordonia:
Grep tool implementation (thanks Sonnet for the help)

goes after #202


debug_gym/gym/tools/grep.py
xingdi-eric-yuan, line 1: May need to integrate the `env._is_ignored_func` here.

sordonia: can you expand?

PR #207, opened 2025-08-01
Refac reasoning content

xingdi-eric-yuan:


debug_gym/agents/debug_agent.py
```diff
@@ -49,7 +49,11 @@ def run(self, task_name=None, debug=False):
                 if debug:
                     breakpoint()
 
-                info = self.env.step(llm_response.tool, llm_response.response)
+                info = self.env.step(
```
sordonia, line 52: why don't we pass the full llm_response object to step?

xingdi-eric-yuan: because it contains other info such as the entire prompt (sys + list of messages), they are not useful in `env.step`

sordonia: ok

PR #209, opened 2025-08-03
Bash tool

sordonia:
Goes after #203

### DIFF
--- a/debug_gym/gym/tools/bash.py
+++ b/debug_gym/gym/tools/bash.py
@@ -1,0 +1,46 @@
+from debug_gym.gym.entities import Observation
+from debug_gym.gym.tools.tool import EnvironmentTool
+from debug_gym.gym.tools.toolbox import Toolbox
+
+
+@Toolbox.register()
+class BashTool(EnvironmentTool):
+    name: str = "bash"
+    examples = [
+        """bash(command="ls -la") to list all files and directories with detailed information.""",
+        """bash(command="grep -r 'function_name' .") to search for 'function_name' in all files recursively.""",
+        """bash(command="find . -name '*.py' | head -10") to find Python files in the current directory.""",
+        """bash(command="cat file.txt | head -20") to show the first 20 lines of a file.""",
+        """bash(command="sed -n 10,25p path/to/file") to show lines 10 to 25 of a file at relative path.""",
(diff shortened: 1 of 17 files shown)


debug_gym/gym/tools/bash.py
```diff
@@ -0,0 +1,50 @@
...
+        """bash(command="ls -la") to list all files and directories with detailed information.""",
+        """bash(command="grep -r 'function_name' .") to search for 'function_name' in all files recursively.""",
+        """bash(command="find . -name '*.py' | head -10") to find Python files in the current directory.""",
+        """bash(command="cat file.txt | head -20") to show the first 20 lines of a file.""",
+        """bash(command="sed -n 10,25p path/to/file") to show lines 10 to 25 of a file at relative path.""",
+        """bash(command="pip list") to show installed Python packages.""",
+    ]
+    description = (
+        "Run commands in a bash shell. "
+        "You have access to common linux and python packages via apt and pip. "
```
MarcCote: We are not running the Docker image as root. So `apt` won't work. We could install `sudo` as part of the docker container configuration if we really need `apt`.

debug_gym/gym/tools/bash.py
```diff
@@ -0,0 +1,50 @@
...
+        "State is persistent across command calls within the same session. "
```
MarcCote, line 20: The current implementation you have below would be stateless, i.e. `terminal.run` will execute a single command and wait for it to finish, then returns the stdout/stderr. If we want to support stateful we need to start a `ShellSession` like we did with the `PdbTool`.

debug_gym/gym/tools/bash.py
sordonia: ok will remove

debug_gym/gym/tools/bash.py
sordonia: why is it stateless? if the agent creates a file `touch file.txt` this file will be present for subsequent calls?

xingdi-eric-yuan: Say if you `cd src`, and next time when you call the bash tool, it's still in the root, outside `src`

sordonia: i see

matheper (approved):

PR #211, opened 2025-08-05
Fix Skip

xingdi-eric-yuan:
Fix a set of bugs in the skipping logic. Now it only skips `"resolved", "skip-resolved", "unresolved", "skip-unresolved"`.

### DIFF
--- a/scripts/run.py
+++ b/scripts/run.py
@@ -62,8 +62,7 @@ def run_agent(args, problem, config):
         if (
             not args.force_all
             and previous_run is not None
-            and previous_run.status
-            in ["resolved", "skip-resolved", "unresolved", "skip-unresolved"]
+            and previous_run.status in ["resolved", "unresolved"]
         ):
             task_logger.debug(f"Previous run found: {problem_path}")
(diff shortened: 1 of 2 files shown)


matheper:
We can change the code to not dump the json status if we skip a problem, and only check for `resolved` and `unresolved`

scripts/run.py
```diff
@@ -58,8 +58,12 @@ def run_agent(args, problem, config):
...
-        previous_run = load_previous_run_status(exp_path, problem)
-        if not args.force_all and previous_run is not None:
+        previous_run = load_previous_run_status(problem_path, problem)
+        if (
+            not args.force_all
+            and previous_run is not None
+            and previous_run.status in ["resolved", "unresolved"]
+        ):
             task_logger.debug(f"Previous run found: {problem_path}")
             success = previous_run.status in ["resolved", "skip-resolved"]
```
matheper:
```suggestion
            success = previous_run.status == "resolved"
```

matheper (approved):

PR #221, opened 2025-08-16
Critique insertion

icwhite:
Enable the capability to give feedback to the model at a given point that has been labeled as a "rollback" trajectory.


debug_gym/agents/debug_agent.py
icwhite, line 1: relevant part of the code to modify the history

PR #234, opened 2025-09-09
Add KubernetesTerminal

MarcCote:
This pull request refactors how terminal classes are imported and introduces support for Kubernetes-based terminals alongside Docker and local terminals. It also adds new terminal implementations and utilities, improving modularity and extensibility for different execution environments. The main changes are grouped below:

**Terminal Import Refactoring and Support Expansion**

* Refactored all environment files (`aider.py`, `env.py`, `mini_nightmare.py`, `r2egym.py`, `swe_bench.py`, `swe_smith.py`) to import terminal classes from the new `debug_gym.gym.terminals` submodules, replacing previous imports from `debug_gym.gym.terminal`. This improves code organization and prepares for multiple terminal backends. [[1]](diffhunk://#diff-0fd30e53ec10b12a89fd527c14c2d8c898af0b24e565ac866877c8c6e322e0f5L10-R11) [[2]](diffhunk://#diff-dd500acaa499f131456c6cd77f9310117f349f19f869fc0fcb4279f71025f108L7-R7) [[3]](diffhunk://#diff-0c9839d137c9c2cedc011109ff42d57f42e819ec862b943a5df9da45f86782f8L8-R9) [[4]](diffhunk://#diff-479feddec077726ab2534d6302d6985811e093be60f684b52240a893391f8e30L24-R26) [[5]](diffhunk://#diff-14c661bab81e165fba55704ca88ccbf72e3a1a1f5c528ff9857527106d3f34ddL13-R15) [[6]](diffhunk://#diff-dc23c08d05edaeab65b40bf7c47817882fcf7469572cb67b26973049578cda7dL18-R18)
* Updated `R2EGymEnv` and `SWEBenchEnv` initialization logic to accept both `DockerTerminal` and the new `KubernetesTerminal`, raising an error for unsupported terminal types. [[1]](diffhunk://#diff-479feddec077726ab2534d6302d6985811e093be60f684b52240a893391f8e30L81-R86) [[2]](diffhunk://#diff-14c661bab81e165fba55704ca88ccbf72e3a1a1f5c528ff9857527106d3f34ddL29-R34)

**New Terminal Implementations and Utilities**

* Added a new `DockerTerminal` class in `docker.py` with methods for running commands, managing containers, copying files, and session handling. This provides a robust Docker-based execution environment.
* Added a new `ShellSession` class in `shell_session.py` for managing shell sessions, including command execution, output reading, and session lifecycle management.
* Implemented a `select_terminal` utility in `terminals/__init__.py` to instantiate the appropriate terminal class (`DockerTerminal`, `KubernetesTerminal`, or `Terminal`) based on configuration, with warnings for Docker-only parameters in local terminals.

These changes collectively improve the modularity and flexibility of the environment setup, allowing for easier extension to new terminal types and better separation of concerns.


debug_gym/gym/terminals/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+from debug_gym.gym.terminals.kubernetes import KubernetesTerminal
+from debug_gym.gym.terminals.terminal import Terminal
+from debug_gym.logger import DebugGymLogger
+
+
+def select_terminal(
+    terminal_config: dict | None = None, logger: DebugGymLogger | None = None
+) -> Terminal | None:
+    if terminal_config is None:
+        return None
```
matheper, line 14: Is there any reason/use case to return None here?

debug_gym/gym/terminals/kubernetes.py
```diff
@@ -0,0 +1,452 @@
...
+                        limits={"memory": "16Gi", "cpu": "2"},
+                        requests={"memory": "1Gi", "cpu": "0.5"},
+                    ),
+                    stdin=True,
+                    stdin_once=False,
+                    tty=True,
+                )
+            ],
+            restart_policy="Never",
+            image_pull_secrets=[{"name": "dockerhub-pro"}],
```
matheper: We probably want this as a config, but I'm not sure how to do that. A base template, perhaps?

debug_gym/gym/terminals/kubernetes.py
```diff
@@ -0,0 +1,452 @@
...
+                    value="true",
+                    effect="NoSchedule",
+                ),
+            ],
+        )
+
+        pod_metadata = client.V1ObjectMeta(
+            name=self.pod_name,
+            namespace=self.namespace,
+            labels={"app": "debug-gym", "component": "terminal"},
```
matheper: Should we add a label with the experiment  uuid?

MarcCote: Sure, we can. Added.

MarcCote: Actually. We don't pass the experiment uuid to it.

scripts/config_r2egym.yaml
```diff
@@ -18,6 +18,25 @@ base:
     }
     terminal: {
         type: "docker",
```
matheper: Should the type be "Kubernetes" here and in config_swebench?

MarcCote: If we want to use kubernetes yes.

matheper: Got it, thanks! If I understand correctly, you want to provide an example in the configs for kubernetes, but I find a bit weird adding these configs if the type is docker

matheper: I don't have any better idea, feel free to resolve this comment 😄

MarcCote: Ah, you say that because of the rest of the settings. You are right. But that was just for debugging. I'm planning to set it back to docker by default.

matheper: Yes, I think docker should be the default. But it would be nice to have the kubernetes configs somewhere.

xingdi-eric-yuan: Yes I agree, the default config should be runnable without modification.

debug_gym/gym/terminals/docker.py
```diff
@@ -153,6 +153,7 @@ def setup_container(self) -> docker.models.containers.Container:
             network_mode="host",
             mem_limit="16G",
         )
+        container.reload()  # Refresh container attributes (e.g., status="running")
```
MarcCote, line 154: That's really needed?

matheper: Yes, otherwise the container status remains "created", even after the container is already running. If that doesn't matter for the logic in docker.py I can change the tests instead.

matheper (approved):

matheper (approved):

PR #240, opened 2025-09-29
allow rewrite to create new files

fverac:
Extends the rewrite tool to be able to create new files.

adds new optional parameter "is_new_file" to the tool.

I believe the ability to create new files with rewrite will be a precursor feature to an agent that is able to write it's own tests (rather than rely on "gold" tests)


xingdi-eric-yuan:
In this PR, we also rename the rewrite tool to the edit tool.

debug_gym/gym/tools/edit.py
```diff
@@ -0,0 +1,165 @@
...
+        path: str = None,
+        start: int = None,
+        end: int = None,
+        new_code: str = "",
+    ) -> Observation:
+        self.edit_success = False
+        if path is None:
+            return self.fail(environment, "File path is None.")
+
+        try:
```
fverac, line 114: worth commenting the purpose here?

Originally there was comment "# If creating a new file, just ensure the target directory is inside workspace and not ignored"

PR #241, opened 2025-09-30
Invert check terminal logic: FORCE_DOCKER_TERMINAL -> ALLOW_LOCAL_TERMINAL

matheper:
This pull request updates the terminal environment checks for bash and grep tools, switching from requiring a Docker terminal to requiring a non-local terminal by default. It also changes the corresponding environment variable from `FORCE_DOCKER_TERMINAL` to `ALLOW_LOCAL_TERMINAL` in both the codebase and CI configuration files.

### DIFF
--- a/debug_gym/gym/tools/bash.py
+++ b/debug_gym/gym/tools/bash.py
@@ -31,19 +31,17 @@ class BashTool(EnvironmentTool):
     def use(self, environment, command: str) -> Observation:
         """Execute a bash command in the environment's terminal and return the result."""
         try:
-            # Assert that the terminal is a Docker terminal (only in production)
+            # Assert that the terminal is not a local terminal (only in production)
             import os
 
-            from debug_gym.gym.terminals.docker import DockerTerminal
+            from debug_gym.gym.terminals.terminal import Terminal
 
-            # Skip Docker terminal check during testing or when explicitly disabled
-            require_docker = (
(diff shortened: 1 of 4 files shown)


debug_gym/gym/tools/bash.py
```diff
@@ -31,19 +31,19 @@ class BashTool(EnvironmentTool):
...
 
-            from debug_gym.gym.terminals.docker import DockerTerminal
+            from debug_gym.gym.terminals.terminal import Terminal
 
-            # Skip Docker terminal check during testing or when explicitly disabled
-            require_docker = (
-                os.getenv("FORCE_DOCKER_TERMINAL", "true").lower() == "true"
+            # Require remote terminal unless local is explicitly allowed
+            require_remote = (
+                os.getenv("ALLOW_LOCAL_TERMINAL", "false").lower() == "false"
```
MarcCote: Really neatpicking but I was wondering which of `os.environ.get` and `os.getenv` was better. Here's what I found.
https://stackoverflow.com/a/55940480

MarcCote (approved):
Should we rename `Terminal` to `LocalTerminal` such that it is clearer what we mean by local?

debug_gym/gym/tools/bash.py
matheper: Interesting! Good to know about `os.putenv`.Done

PR #247, opened 2025-10-02
Chunk very long commands when writing file.

### LINKED ISSUE #245: Deal with long bash command
@xingdi-eric-yuan  reported that long command line were failing (e.g., cat huge content into a file).

The proper way should be to break command in multiple smaller command. Easy to do for `cat` not sure how to deal with that for agent's bash commands.

MarcCote:
This pull request introduces improvements to file writing in the workspace and enhances the logger's debug mode activation. The most significant changes are the addition of chunked file writing to handle large contents, more robust detection of debug mode via environment variables, and expanded test coverage for edge cases in file writing.

**Workspace file writing improvements:**

* The `write_file` method in `debug_gym/gym/workspace.py` now splits content into 32kB chunks when writing to files, ensuring compatibility with shell command length limits and supporting very large files.

**Logger enhancements:**

* The debug mode in `debug_gym/logger.py` is now activated if the `DEBUG_GYM_DEBUG` environment variable matches common truthy values like `"1"`, `"true"`, `"yes"`, or `"on"`, making it more flexible and user-friendly.

**Test coverage improvements:**

* The test `test_write_file` in `tests/gym/test_workspace.py` now includes checks for writing empty content and very large content (2MB), verifying that the new chunked writing logic works correctly for these edge cases.

Closes #245

### DIFF
--- a/debug_gym/gym/workspace.py
+++ b/debug_gym/gym/workspace.py
@@ -136,14 +136,24 @@ class Workspace:
         """Writes `content` to `filepath` exactly as-is, preserving any trailing newlines."""
         abs_filepath = self.resolve_path(filepath)
 
+        # We will split content in chunks of 32kB to avoid hitting command length limits.
+        chunk_size = 32 * 1024  # 32kB
+        first_chunk = content[:chunk_size]
+        rest = content[chunk_size:]
+
         # In the following command we:
         # - use a single-quoted heredoc (cat <<'nDEBUGGYM_EOF' ... nDEBUGGYM_EOF) so the heredoc body is taken literally (no shell expansion)
         # - append a sentinel character DEBUGGYM_DEL inside the heredoc so we can detect/restore trailing newlines later
         # - capture the heredoc output into shell variable CONTENT since command substitution strips trailing newlines
         # - "${CONTENT%DEBUGGYM_DEL}" removes the trailing sentinel DEBUGGYM_DEL (restoring the original trailing-newline state)
         # - echo -n writes the result without adding an extra newline
-        cmd = f"CONTENT=$(cat <<'DEBUGGYM_EOF'\n{content}DEBUGGYM_DEL\nDEBUGGYM_EOF\n); echo -n \"${{CONTENT%DEBUGGYM_DEL}}\" > {abs_filepath}"
+        cmd = f"CONTENT=$(cat <<'DEBUGGYM_EOF'\n{first_chunk}DEBUGGYM_DEL\nDEBUGGYM_EOF\n); echo -n \"${{CONTENT%DEBUGGYM_DEL}}\" > {abs_filepath}"
         self.terminal.run(cmd, raises=True)
+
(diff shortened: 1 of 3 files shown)


xingdi-eric-yuan (approved):

tests/gym/test_workspace.py
```diff
@@ -312,3 +312,13 @@ def test_write_file(workspace):
     file_content_with_newlines = "Hello, DebugGym!\nThis is a test.\n"
     workspace.write_file("test.txt", file_content_with_newlines)
     assert file_path.read_text() == file_content_with_newlines
+
+    # Test with empty content
```
matheper: Nitpicking here, up to you to change this or not. If creating the workspace fixture isn't too expensive, I would split each block into their own test. It's easy to identify and fix bugs when tests are shorter.

debug_gym/gym/workspace.py
```diff
@@ -136,15 +136,25 @@ def write_file(self, filepath: str, content: str):
...
+        rest = content[chunk_size:]
+
         # In the following command we:
         # - use a single-quoted heredoc (cat <<'nDEBUGGYM_EOF' ... nDEBUGGYM_EOF) so the heredoc body is taken literally (no shell expansion)
         # - append a sentinel character DEBUGGYM_DEL inside the heredoc so we can detect/restore trailing newlines later
         # - capture the heredoc output into shell variable CONTENT since command substitution strips trailing newlines
         # - "${CONTENT%DEBUGGYM_DEL}" removes the trailing sentinel DEBUGGYM_DEL (restoring the original trailing-newline state)
         # - echo -n writes the result without adding an extra newline
-        cmd = f"CONTENT=$(cat <<'DEBUGGYM_EOF'\n{content}DEBUGGYM_DEL\nDEBUGGYM_EOF\n); echo -n \"${{CONTENT%DEBUGGYM_DEL}}\" > {abs_filepath}"
+        cmd = f"CONTENT=$(cat <<'DEBUGGYM_EOF'\n{first_chunk}DEBUGGYM_DEL\nDEBUGGYM_EOF\n); echo -n \"${{CONTENT%DEBUGGYM_DEL}}\" > {abs_filepath}"
```
matheper, line 150: Is the different of this block compared to the loop, the single ">"? Can you just create the empty cmd here and remove this first-chunk logic?

matheper: nvm, got it now

tests/gym/test_workspace.py
MarcCote: True. Good idea, it is using the LocalTerminal so it's fast.

debug_gym/gym/workspace.py
MarcCote: Yes, but that would mean a minimum of two calls to terminal.run each time. As I implemented it, only when the content is too large, that we will have 1+ calls.

matheper: Oh, I see! That makes sense!

Issue #249, opened 2025-10-05
Create a HuggingFaceLLM


xingdi-eric-yuan:
Create a HuggingFaceLLM rather than just use OpenAILLM for vLLM models, in such way we can provide HF's tokenizer. Currently we are just using tiktokens.

PR #251, opened 2025-10-05
HuggingFace LLM

xingdi-eric-yuan:
This is to address https://github.com/microsoft/debug-gym/issues/249

### DIFF
--- a/debug_gym/llms/base.py
+++ b/debug_gym/llms/base.py
@@ -66,6 +66,8 @@ class LLMConfig:
     scope: Optional[str] = None
     # Custom parameters to pass to generate
     generate_kwargs: dict = None
+    # Additional kwargs for tokenizer construction (e.g., trust_remote_code)
+    tokenizer_kwargs: dict | None = None
 
     def __post_init__(self):
         # Set tokenizer to model if not specified
@@ -78,6 +80,8 @@ class LLMConfig:
             self.tags = []
         if self.generate_kwargs is None:
             self.generate_kwargs = {}
+        if self.tokenizer_kwargs is None:
+            self.tokenizer_kwargs = {}
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


debug_gym/llms/base.py
```diff
@@ -277,6 +285,34 @@ def count_tokens(self, text: str) -> int:
...
+
+        Args:
+            messages: Original message list.
+            message_token_counts: Optional precomputed counts aligned with messages.
+
+        Returns:
+            A trimmed list of messages.
+        """
+
+        from debug_gym.agents.utils import trim_prompt_messages
```
MarcCote: why the local import, here?

debug_gym/llms/base.py
```diff
@@ -277,6 +285,34 @@ def count_tokens(self, text: str) -> int:
...
         return len(self.tokenize(text))
 
+    def _get_message_token_counts(self, messages: list[dict]) -> list[int]:
+        """Return per-message token counts used for context management.
+
+        Subclasses can override this to plug in custom counting strategies
+        (for example, chat-template aware tokenizers).
+        """
+
+        from debug_gym.agents.utils import get_message_tokens
```
MarcCote: and here

debug_gym/llms/base.py
matheper: I believe it's because it causes a circular import otherwise.

matheper: We could solve this by removing the imports from the `__init__.py` files, for example, in `llms` and `agents`

matheper: I removed the imports from agents `__init__` and kept the ones from llms so we can still call
`from debug_gym.llms import LLMCLASS`

matheper (approved):

PR #253, opened 2025-10-17
Fverac/swebench self test agent

fverac:


debug_gym/agents/base_agent.py
```diff
@@ -266,6 +266,24 @@ def run(self, task_name=None, debug=False):
...
+                          ================================================================
+                            ================================================================
+                            ================================================================
+                            ================================================================
+                            ================================================================
+                          ATTENTION ATTENTION ATTENTION ATTENTION ATTENTION ATTENTION
+                          submit cool called
+                    """
+                    )
+                    # break
```
MarcCote: I think the submit tool should set env.done = True. (I did that change already). This means we don't have to break here.

debug_gym/agents/swe_agent.py
```diff
@@ -0,0 +1,7 @@
+from debug_gym.agents.base_agent import BaseAgent, register_agent
+
+
+@register_agent
+class SWEAgent(BaseAgent):
+    name = "swe_agent"
```
MarcCote, line 6: Decided to call it swe_agent instead of bash_agent.

debug_gym/gym/envs/env.py
```diff
@@ -338,22 +340,25 @@ def reset(self, *, options: dict = None):
         self.queue_event(Event.ENV_RESET, source="env")
         self.all_observations = self.process_events()
 
-        # Gets eval (initial observation) from cache or by running env.eval
-        if self.last_eval:  # if eval tool was triggered by Event.ENV_RESET
-            self.step_observation = Observation("env", self.last_eval.output)
-        else:  # if eval tool was not triggered by Event.ENV_RESET
+        # Gets eval (initial observation) by running env.eval if debug_mode is on.
+        eval_observation = None
+        self.step_observation = Observation("env", self.instructions)
```
MarcCote, line 344: When debug_mode is false, I'm setting the first observation as the instructions. However it is a bit redundant with the system prompt for the current agent implementation we have. @xingdi-eric-yuan

debug_gym/agents/base_agent.py
MarcCote: Also, this raises another point. In debug-gym we assume that when `env.info` is True then it was successful. 

I think this is wrong and we should determine success by looking at the `score == max_score`. Actually we should renamed it to `terminated` as in https://gymnasium.farama.org/api/env/#gymnasium.Env.step 
@matheper

debug_gym/gym/envs/swe_bench.py
```diff
@@ -132,6 +132,8 @@ def setup_task(self, task_name: str, options: dict = None):
         self.entrypoint = self.entrypoint.replace("--tb=no", "--tb=short")
 
         self.git_apply_cmd = f"git apply -"
+        # Flag to ensure final submission only happens once.
+        self._final_submitted = False
```
MarcCote: not needed

debug_gym/gym/envs/swe_bench.py
```diff
@@ -186,6 +189,47 @@ def apply_gold_patch(self):
         self.terminal.run(command, raises=True)
         self.logger.info("Patch applied successfully.")
 
+    def eval(self, **kwargs) -> EvalOutput:
```
MarcCote, line 183: I moved your final_submit logic directly into the eval function and made the submit tool call it instead.

debug_gym/gym/tools/eval.py
```diff
@@ -13,9 +13,9 @@ def use(self, environment) -> Observation:
         eval_output = environment.eval()
         return Observation(self.name, eval_output.output)
 
-    def on_env_reset(self, environment, **kwargs):
-        super().on_env_reset(environment, **kwargs)
-        return self(environment)
+    # def on_env_reset(self, environment, **kwargs):
+    #     super().on_env_reset(environment, **kwargs)
+    #     return self(environment)
```
MarcCote: @matheper I decided to remove this and always call the eval in the reset (if debug_mode if true).

scripts/config_swebench.yaml
```diff
@@ -45,3 +45,12 @@ debug_5_agent:
...
     tools: ["eval", "pdb"]
+
+swe_agent:
+    env_kwargs:
+        debug_mode: false
+    
+    max_steps: 100
+    max_rewrite_steps: 20
+    tools: ["bash", "rewrite", "submit"]
+    #tools: ["bash", "pdb", "rewrite", "submit"]
```
MarcCote: Right now, it is unclear what entrypoint for pdb should be used. I think we need to move towards the agent providing the entrypoint to the tool.  cc @xingdi-eric-yuan

MarcCote: @fverac had a workaround for it: calling `python -m pdb -m pytest .`, i.e. running all the tests in the repo. Could work but will take a long time to run for some of the repo. I prefer the agent to pass the entry point to the pdb tool.

debug_gym/gym/envs/env.py
xingdi-eric-yuan: maybe in that mode, we don't need a "first observation"?

MarcCote: Could be an option. We should have a different type of agent, actually. Not just subclassing base_agent. So we could have a swe-agent that mimics the original swe-agent (maybe the mini version).

debug_gym/agents/base_agent.py
matheper: Is there any property in `gymnasium.Env` with a reason to why it is not terminated (error, max_steps, etc)?

matheper: I see, there's `terminated` and `truncated`.

debug_gym/gym/tools/submit.py
```diff
@@ -0,0 +1,18 @@
+from debug_gym.gym.entities import Observation
+from debug_gym.gym.tools.tool import EnvironmentTool
+from debug_gym.gym.tools.toolbox import Toolbox
+
+
+@Toolbox.register()
+class SubmitTool(EnvironmentTool):
+    name = "submit"
+    description = (
+        "Finalize the task: applies the hidden official benchmark test patch,"
```
matheper: This is the instruction given to the LLM when selecting a tool to call, right?
I'm not sure we want to say this is part of a benchmark or just have a more generic tool description.

debug_gym/gym/tools/eval.py
MarcCote: Putting it back

debug_gym/gym/envs/env.py
```diff
@@ -338,30 +340,31 @@ def reset(self, *, options: dict = None):
...
             current_breakpoints=self.current_breakpoints(),
             action_reasoning=None,
             action_content=None,
             action_tool_call=None,
-            done=self.done,
+            terminated=self.terminated,
+            resolved=self.resolved,
             score=self.score,
-            max_score=self.max_score,
+            max_score=self.max_score or 1,
```
matheper: Can you make this a @property?

matheper: Or, is this the same as calculate_max_score?

debug_gym/gym/envs/env.py
```diff
@@ -383,11 +386,16 @@ def calculate_score(self, eval_output: EvalOutput) -> int:
         Override in subclasses for different behavior."""
         return eval_output.success
 
-    def calculate_done(self, eval_output: EvalOutput) -> bool:
-        """Determine if the task is done.
+    def calculate_resolved(self, eval_output: EvalOutput) -> bool:
+        """Determine if the task has been resolved.
         Override in subclasses for different behavior."""
         return self.score == self.max_score
```
matheper, line 391: Same here, should this include "or 1"

debug_gym/gym/envs/env.py
MarcCote: yes, but calculate max score requires a proper self.last_eval which is only set after we have called the eval() once. To be fair, I don't like this assumption of max_score can only be obtained after running eval once but we need to know how many tests there are.

debug_gym/gym/envs/env.py
MarcCote: ah maybe that's the way to do it.

debug_gym/gym/envs/env.py
```diff
@@ -464,22 +472,28 @@ def step(
         self.all_observations.insert(0, self.step_observation)
 
         # Calculate score and done based on the last eval output
-        self.score = self.calculate_score(self.last_eval)
-        self.done = self.calculate_done(self.last_eval)
+        if self.last_eval:
+            self.max_score = self.max_score or self.calculate_max_score(self.last_eval)
```
matheper: Why setting `self.max_score` here?

debug_gym/gym/envs/env.py
```diff
@@ -464,22 +472,28 @@ def step(
...
             dir_tree=self.workspace.display_files(self.dir_tree_depth),
             current_breakpoints=self.current_breakpoints(),
             action_reasoning=action_reasoning,
             action_content=action_content,
             action_tool_call=action_tool_call,
             instructions=self.instructions,
             score=self.score,
-            max_score=self.max_score,
-            done=self.done,
+            max_score=self.max_score or 1,
```
matheper: property?

debug_gym/gym/tools/eval.py
```diff
@@ -18,6 +20,7 @@ def on_env_reset(self, environment, **kwargs):
         return self(environment)
 
     def on_rewrite_success(self, environment, **kwargs):
+        # TODO: Make this behavior configurable via tool arguments
```
matheper: Are you addressing this now or is it for future ourselves?

tests/gym/tools/test_pdb.py
```diff
@@ -693,9 +694,10 @@ def test_pdb_list_output_indentation(tmp_path, setup_pdb_repo_env):
         f.write("\n".join(f"    'Line {i+1}'" for i in range(1, 2000)))
         f.write("\n\nif __name__ == '__main__':\n")
         f.write("    dummy_function()\n")
-    env.set_entrypoints("python large_file.py", "python -m pdb large_file.py")
-    pdb_tool.start_pdb(env)
-    pdb_obs = pdb_tool.use(env, "b large_file.py:100")
+    # env.set_entrypoints("python large_file.py", "python -m pdb large_file.py")
```
matheper: Remove unused code?

debug_gym/gym/envs/env.py
```diff
@@ -25,7 +25,8 @@ class EnvInfo:
     instructions: dict
     score: int
     max_score: int
-    done: bool
+    terminated: bool
```
matheper: Can you add a short inline comment to `terminated` and `resolved` properties, just to make clear what the differences are.

debug_gym/gym/envs/env.py
MarcCote: I don't like the current logic, if you have suggestion let me know. Right now, the rational is when we run the swe-agent version, the eval has not been called (i.e. we don't apply the gold tests just yet). But whenever we self.last_eval becomes not none, we should compute the actual max_score. And we only do it if max_score was not already computed.

debug_gym/gym/tools/eval.py
MarcCote: This is for the future. Will make a Github issue instead.

- [x] Make Github issue about making the behavior of the eval tool configurable in the yaml file. E.g. eval_on_rewrite=True/False. This will remove it from Env().

tests/gym/tools/test_pdb.py
MarcCote: Good catch.

debug_gym/gym/envs/env.py
fverac: I'm not sure I understand the suggestion.

You mean doing `self.score == (self.max_score or 1)`?

fverac:
@matheper ready for another look

Issue #266, opened 2025-10-31
Make the eval tool parametrizable


MarcCote:
Let's make the `eval` tool accept `eval_on_rewrite=True/False` as optional parameters when defining the list of tools to use in the YAML config.

PR #269, opened 2025-11-05
make eval tool parameterizable

fverac:
RE: #266 

Adds auto_eval_on_rewrite init param to eval. If present, will take priority over env auto_eval_on_rewrite arg.

### DIFF
--- a/tests/gym/tools/test_pdb.py
+++ b/tests/gym/tools/test_pdb.py
@@ -197,10 +197,13 @@ def test_pdb_use_docker_terminal(tmp_path, setup_test_repo):
     terminal = DockerTerminal(
         base_image="python:3.12-slim",
         setup_commands=["apt update", "apt install -y git tree", "pip install pytest"],
-        env_vars={"PYTHONDONTWRITEBYTECODE": "1"},  # avoid __pycache__
+        env_vars={
+            "PYTHONDONTWRITEBYTECODE": "1",  # avoid __pycache__
+            "PYTEST_DISABLE_PLUGIN_AUTOLOAD": "1",  # disable plugins that might add escape sequences
+        },
     )
-    # no:cacheprovider to avoid .pytest_cache
-    debug_entrypoint = "python -m pdb -m pytest -p no:cacheprovider -sv ."
+    # no:cacheprovider to avoid .pytest_cache, --tb=short to reduce output
+    debug_entrypoint = "python -m pdb -m pytest -p no:cacheprovider --color=no -sv ."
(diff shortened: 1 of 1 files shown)


MarcCote:
I agree. Let's just remove it.

debug_gym/agents/base_agent.py
```diff
@@ -71,7 +71,10 @@ def parse_reasoning_model_response(self, response, reasoning_end_token):
 
     def _auto_eval_on_rewrite(self):
         """Check if auto eval on rewrite is enabled."""
-        return self.config.get("env_kwargs", {}).get("auto_eval_on_rewrite", False)
+        try:
+            return self.env.get_tool("eval").auto_eval_on_rewrite
+        except:
```
MarcCote: Let's specify the exact error type in the `except`.

debug_gym/gym/envs/env.py
```diff
@@ -241,6 +239,13 @@ def __init__(
         self.rng = None
         self.additional_kwargs = kwargs
 
+        if "auto_eval_on_rewrite" in kwargs:
+            raise ValueError(
+                "The 'auto_eval_on_rewrite' parameter is no longer supported. "
+                "Please remove it from your initialization arguments."
+                "Instead, set 'auto_eval_on_rewrite' in the EvalTool instance."
```
MarcCote, line 246: Should we mention the config YAML instead?

tests/gym/tools/test_eval.py
```diff
@@ -60,3 +59,22 @@ def test_eval_on_event(env, method, env_auto_eval_on_rewrite, expected):
...
+
+    eval_call = ToolCall(id="eval_id", name="eval", arguments={})
+    env_info = env.step(eval_call)
+    assert "FAILED test_1.py::test_1" in env_info.step_observation.observation
+
+    # Edit test file to pass
+    with open(env.working_dir / "test_1.py", "w") as f:
+        f.write("def test_1():\n  assert True\n")
+
+    # Tool's None should defer to env's True setting
```
MarcCote: What does this means?

debug_gym/gym/envs/env.py
MarcCote: Or maybe not. It's more accurate to talk about EvalTool instance, but from a user point of view, it might be unclear where to set it. Maybe we can also show an example of turning it on.

tests/gym/tools/test_eval.py
fverac: outdated comment, fixing!

Issue #281, opened 2025-11-26
Use AgentArgs dataclass instead of config dict


MarcCote:
Using a dataclass to provide settings to Agent init would be cleaner.

PR #287, opened 2025-11-27
Agent args dataclass

sordonia:
Asked codex for #281

Goes after #285

### DIFF
--- a/debug_gym/gym/envs/env.py
+++ b/debug_gym/gym/envs/env.py
@@ -224,6 +224,8 @@ class RepoEnv(TooledEnv):
         self.infos: EnvInfo | None = None
         self.rng = None
         self.additional_kwargs = kwargs
+        self.task_name: str | None = None
+        self.options: dict = {}
 
         if "auto_eval_on_rewrite" in kwargs:
             raise ValueError(
@@ -320,11 +322,11 @@ class RepoEnv(TooledEnv):
 
     def reset(self, *, options: dict = None):
         """Resets the environment and returns eval as the initial observation."""
-        options = options or {}
+        self.options = options if options is not None else self.options
         self.logger.debug("Resetting environment")
         self.close()  # Clean up previous workspace and terminal.
-        self.task_name = options.get("task_name")
-        self.setup_task(task_name=self.task_name, options=options)
+        self.task_name = self.options.get("task_name")
+        self.setup_task(task_name=self.task_name, options=self.options)
         self.setup_workspace()
         self.setup_terminal()
         self._reset_env_state()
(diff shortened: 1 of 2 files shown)


debug_gym/agents/base_agent.py
```diff
@@ -17,6 +18,71 @@
...
+@dataclass
+class AgentArgs:
+    random_seed: int
+    memory_size: int
+    max_steps: int
+    max_rewrite_steps: int
+    env_kwargs: Dict[str, Any] = field(default_factory=dict)
+    reset_prompt_history_after_rewrite: bool = False
+    system_prompt_template_file: str | None = None
+    n_rewrites_before_pdb: int = 0
```
MarcCote: That settings is only used by the debug_after_rewrite agent.

debug_gym/agents/base_agent.py
```diff
@@ -33,22 +99,21 @@ class BaseAgent:
...
+        self.args = AgentArgs.from_dict(args) if isinstance(args, dict) else args
         self.env = env
         self.logger = logger or DebugGymLogger("debug-gym")
         self.llm = llm
-        self._uuid = self.config.get("uuid", str(uuid.uuid4()))
-        self._output_path = pjoin(self.config["output_path"], self._uuid)
+        self._uuid = self.args.uuid
 
-        os.makedirs(self._output_path, exist_ok=True)
+        self.env.dir_tree_depth = self.args.show_directory_tree
```
sordonia: this is still weird

debug_gym/agents/base_agent.py
```diff
@@ -176,12 +233,12 @@ def _default_system_prompt(self, info) -> str:
             "Instructions": info.instructions,
         }
 
-        if self._show_directory_tree():
+        if self.args.show_directory_tree > 0:
```
sordonia: all these options into the 'base agent' seem quite overkill

debug_gym/agents/base_agent.py
```diff
@@ -33,22 +99,21 @@ class BaseAgent:
 
     def __init__(
         self,
-        config: dict,
+        args: AgentArgs | Dict[str, Any],
         env: RepoEnv,
```
sordonia: can we remove dependence on env?

debug_gym/agents/base_agent.py
MarcCote: We can have a Froggy agent, those are the settings used for the tech report.

debug_gym/agents/base_agent.py
MarcCote: r2egym agent does the same no? We get a free list dir as the first observation

sordonia: no r2e agent issues usually a ls at the beginning

debug_gym/agents/base_agent.py
sordonia: yes

MarcCote: Made a FroggyAgent and stripped those configs from BaseAgent.

PR #291, opened 2025-11-27
Envs for images

sordonia:
also removes dependency on swesmith: #292

PR #294, opened 2025-11-29
Simpler agent

sordonia:
goes after #291

### DIFF
--- a/debug_gym/gym/utils.py
+++ b/debug_gym/gym/utils.py
@@ -127,29 +127,6 @@ def create_ignore_file(
         f.write("\n".join(patterns + [path.name]))
 
 
-def _walk(path, depth: int | None = None, skip: Callable | None = None):
-    """recursively list files and directories up to a certain depth"""
-    depth = 1e5 if depth is None else depth
-    if depth <= 0:  # stop recursion
-        return
-
-    with os.scandir(path) as p:
-        for entry in sorted(p, key=lambda x: x.name):
-            if skip and skip(entry.path):
-                continue
-
-            yield Path(entry)
-            if entry.is_dir() and depth > 0:
-                yield from _walk(entry.path, depth=depth - 1, skip=skip)
-
-
-def is_subdirectory(path, directory):
-    directory = str(directory)
-    if not path.startswith(directory):
-        path = pjoin(directory, path)
-    return not os.path.relpath(path, directory).startswith("..")
-
-
 def cleanup_pytest_output(output):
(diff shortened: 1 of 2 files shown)


debug_gym/agents/froggy_agent.py
```diff
@@ -108,3 +159,25 @@ def _default_system_prompt(self, info) -> str:
             system_prompt_dict["Shortcut features"] = shortcut_features
 
         return self.to_pretty_json(system_prompt_dict)
+
+    def apply_patch(self, patch_path: str) -> bool:
```
MarcCote: Do we even need that?

scripts/run.py
```diff
@@ -114,29 +106,13 @@ def run_agent(args, task_name: str, task_data: dict, config: dict):
...
                 f"the time limit of {args.timeout} seconds."
             )
-            task_logger.report_progress(
-                problem_id=task_name,
-                step=1,
-                total_steps=1,
-                score=0,
-                max_score=None,
-                status="error",
-            )
```
MarcCote, line 136: I think that if you remove those, then the progress bar will be broken, they will stay in the "running" state.

debug_gym/agents/base_agent.py
```diff
@@ -166,58 +152,144 @@ def _load_system_prompt_template(self) -> Template | None:
...
+        )
+        if system_prompt_template:
             # Add custom filter to Jinja2 environment
             env = Environment()
             env.filters["to_pretty_json"] = self.to_pretty_json
             env.filters["trim_message"] = self.trim_message
             return env.from_string(system_prompt_template)
         return None
 
+    def _load_instance_prompt_template(self) -> Template | None:
```
xingdi-eric-yuan: Small refactor to have load_sys_template and load_instance_template calling same helper function?

debug_gym/agents/froggy_agent.py
```diff
@@ -1,12 +1,56 @@
...
+from debug_gym.gym.utils import filter_non_utf8
+from debug_gym.llms.utils import trim
+
+
+def build_history_prompt(
+    history: HistoryTracker, llm: LLM, reset_prompt_history_after_rewrite: bool = False
+):
+    env_observations, llm_responses = history.get()
+    latest_rewrite_step = 0
+    # Find the latest rewrite step if reset_prompt_history_after_rewrite
```
xingdi-eric-yuan: I think this reset_prompt_history_after_rewrite shortcut can be removed?

PR #284, opened 2025-11-26
Free Env

xingdi-eric-yuan:
This pull request introduces a new "FreeEnv" environment and "FreeAgent" agent, enabling open-ended, container-based exploration sessions in debug_gym. It adds support for running arbitrary container images with customized setup and toolchains, including human-in-the-loop operation. The changes also include configuration and tooling updates to support flexible session management and error handling.

**New Environment and Agent:**

* Added `FreeEnv`, a lightweight environment for running arbitrary container images, with support for custom instructions, workspace directories, setup commands, and optional git initialization. (`debug_gym/gym/envs/free_env.py`)
* Introduced `FreeAgent`, a minimal reasoning agent designed for FreeEnv sessions, including improved error handling for environment startup issues. (`debug_gym/agents/free_agent.py`)

**Integration and Selection:**

* Registered `FreeEnv` and `FreeAgent` in their respective module initializers, and updated environment selection logic to support the new "free" type. (`debug_gym/gym/envs/__init__.py`, `debug_gym/agents/__init__.py`) [[1]](diffhunk://#diff-15a1d6c0ac1a6a3e7a5c1e038b9f280528d615077de78f9e69ca94bd63fb1cf4R3) [[2]](diffhunk://#diff-15a1d6c0ac1a6a3e7a5c1e038b9f280528d615077de78f9e69ca94bd63fb1cf4R27-R28) [[3]](diffhunk://#diff-d0fa37847e837ef03bdbb361b64e8347d1b2268ee0fa41180d5860d26cca3d58R1-R3)

**Tooling and Session Management:**

* Updated the `SubmitTool` to allow disabling automatic evaluation upon submission, facilitating open-ended sessions. (`debug_gym/gym/tools/submit.py`)
* Added a configuration file and interactive demo script for FreeEnv sessions, supporting both automated and human-driven workflows. (`scripts/config_free_env.yaml`, `scripts/free_env_human.py`) [[1]](diffhunk://#diff-5081145e729d07f99866d12186215fdd1e9683313aa9f2084864ee37cb793c7fR1-R21) [[2]](diffhunk://#diff-fd3893e16637a32f214d6b2af47a13c50b735a22458083610da25474786dba69R1-R200)

**BaseAgent Improvements:**

[...]

PR #311, opened 2025-12-10
Error Handling

xingdi-eric-yuan:
- Terminate the episode when encountering non-recoverable errors (e.g., pod dead).
- For errors such as tool misuse, do not terminate, pass the error into the agent's next observation, so it can adjust and correct its tool calling action. 
- Every tool now can have their own setup commands. This prevents the termianl to install unnecesary dependencies when a tool is not being added into the toolbox. 

After https://github.com/microsoft/debug-gym/pull/284

### DIFF
--- a/tests/gym/terminals/test_kubernetes.py
+++ b/tests/gym/terminals/test_kubernetes.py
@@ -253,7 +253,7 @@ def test_kubernetes_terminal_session(tmp_path):
 
 @if_kubernetes_available
 def test_copy_content(tmp_path):
-    # Create a temporary source file
+    # Create a temporary source directory with a file
     source_dir = tmp_path / "source_dir"
     source_dir.mkdir()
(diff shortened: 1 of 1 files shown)


tests/test_experiment.py
```diff
@@ -26,14 +26,104 @@ class Args:
...
         logger = DebugGymLogger("test")
 
-        # Create a test repository
+        config = {
+            "terminal": {"type": "docker"},
+            "uuid": "test-uuid",
+            "env": {"run_timeout": 300},
+        }
+        task_data = {"env_type": "swebench", "instance_id": "test-123"}
+
```
MarcCote: Why all the mocking? Github action do have docker installed.

xingdi-eric-yuan: removed the mockings

PR #315, opened 2025-12-13
step / init / execute_action

sordonia:
some cleanup:

- step, init, execution_action methods in agent.py
- llm is required for BaseAgent class
- removed llm as arg to agent.run()


debug_gym/agents/base_agent.py
```diff
@@ -27,8 +27,8 @@ def register_agent(cls):
 
 @dataclass
 class AgentArgs:
-    system_prompt: str | None = None
-    instance_prompt: str | None = None
+    system_prompt: str = ""
+    instance_prompt: str = "Instructions: {{ info.instructions }}"
```
MarcCote, line 31: Nice

PR #352, opened 2026-02-28
Handle response without tool.

MarcCote:


debug_gym/agents/simple_agent.py
```diff
@@ -155,11 +178,24 @@ def parse_tool_call(self, tool_call: str) -> List[ToolCall]:
...
+                    tool_arg_schema = tool.arguments
+
             params = {}
             for param_key, param_value in param_matches:
                 param_key = param_key.strip()
                 param_value = param_value.strip()
-                params[param_key] = param_value
+                params[param_key] = self._cast_param(
+                    param_value, tool_arg_schema.get(param_key)
+                )
```
MarcCote, line 198: Cast the param value according to its type.

debug_gym/gym/envs/swe_bench.py
```diff
@@ -139,19 +139,34 @@ def apply_gold_patch(self):
...
+                        files.append(path)
+        return files
+
     def eval(self, **kwargs) -> EvalOutput:
         # We need to apply the test patch before running any evaluation.
-        # Reset any changes made to test_directives files.
-        self.terminal.run(f"git checkout -- {' '.join(self.test_directives)}")
+        # Reset any changes made to test patch files.
+        test_files = self._extract_files_from_patch(self.test_patch)
+        self.terminal.run(f"git checkout -- {' '.join(test_files)}")
```
MarcCote, line 160: Important change. For some swebench tasks, the test_directives didn't contain the test file name. We need to extract them from the test_patch.

debug_gym/gym/tools/view.py
```diff
@@ -74,6 +74,16 @@ def use(
...
+        # Cast start and end to integers if they are strings
+        try:
+            start = int(start) if start is not None else None
+            end = int(end) if end is not None else None
+        except ValueError:
+            return Observation(
+                self.name,
+                "Invalid line number. Start and end line numbers should be integers.",
+            )
+
```
MarcCote, line 86: I added this before doing the casting, so a bit redundant now, but can't hurt. We know those values should be `int` or None.

debug_gym/llms/human.py
```diff
@@ -504,16 +504,26 @@ def define_tools(self, tool_call_list: list[EnvironmentTool]) -> list[dict]:
...
                     "type": "tool_use",
                     "id": response.tool.id,
                     "name": response.tool.name,
                     "input": response.tool.arguments,
                 }
-            ],
+            )
+        return {
+            "role": "assistant",
+            "content": content,
```
MarcCote, line 526: Adding this check, because some agent doesn't have tool call in their response.

MarcCote (commented):
Added a few comments on the most important parts. @matheper

matheper (approved):

debug_gym/gym/terminals/kubernetes.py
```diff
@@ -367,10 +367,7 @@ def __init__(
 
     @property
     def pod_name(self):
-        if self._pod is None:
-            raise ValueError("Pod not created yet; pod_name is not available.")
-
-        return self._pod.name
+        return self.pod.name
```
matheper, line 370: Thanks! I fixed that locally but never committed it.

debug_gym/gym/envs/swe_bench.py
matheper: Does that mean some tests were never reset?

MarcCote: Yep.

debug_gym/gym/terminals/kubernetes.py
MarcCote: I actually merge your stuff to this branch :D

PR #353, opened 2026-03-03
Configurable Timeout Argument to BashTool

xingdi-eric-yuan:
- LLM can now specify timeout per command: bash(command="pytest", timeout=600)
- Default timeout: 30s (unchanged from previous behavior)
- Maximum timeout: 3600s (1 hour), values above are clamped
- timeout=null uses the default
- Added 4 new tests for timeout behavior

### DIFF
--- a/debug_gym/gym/tools/bash.py
+++ b/debug_gym/gym/tools/bash.py
@@ -11,13 +11,13 @@ class BashTool(EnvironmentTool):
     MAX_TIMEOUT = 3600  # 1 hour
 
     examples = [
-        """bash(command="ls -la") to list all files and directories with detailed information.""",
-        """bash(command="grep -r 'function_name' .") to search for 'function_name' in all files recursively.""",
-        """bash(command="find . -name '*.py' | head -10") to find Python files in the current directory.""",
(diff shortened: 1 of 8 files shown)


debug_gym/gym/tools/bash.py
```diff
@@ -7,13 +7,17 @@
...
+    MAX_TIMEOUT = 3600  # 1 hour
+
     examples = [
         """bash(command="ls -la") to list all files and directories with detailed information.""",
         """bash(command="grep -r 'function_name' .") to search for 'function_name' in all files recursively.""",
         """bash(command="find . -name '*.py' | head -10") to find Python files in the current directory.""",
         """bash(command="cat file.txt | head -20") to show the first 20 lines of a file.""",
         """bash(command="sed -n 10,25p path/to/file") to show lines 10 to 25 of a file at relative path.""",
         """bash(command="pip list") to show installed Python packages.""",
+        """bash(command="python -m pytest tests/ -x", timeout=600) to run tests with a 10-minute timeout.""",
```
MarcCote: I believe those examples are misleading to the model that uses different tool call format, e.g. xml parser.

MarcCote (approved):

MarcCote (approved):

PR #375, opened 2026-08-24
Remove local terminal backend

matheper:
## Summary
- remove the production `LocalTerminal` backend and reject legacy `terminal.type: local` configs with clear Docker/Kubernetes guidance
- keep `LocalEnv` for local-path tasks while requiring an isolated terminal
- run local-path environment and workspace tests in Docker instead of providing a host-backed test terminal
- remove host-filesystem branching from the terminal and workspace abstractions
- lazily load environment and LLM modules to reduce optional dependency coupling

## Validation
- `301 passed` in the focused terminal, workspace, environment, and tool suite
- Black, isort, and `git diff --check` pass


MarcCote (approved):

MarcCote:
@copilot look at the failing tests.

PR #376, opened 2026-08-27
Fix command injection in grep and workspace writes

matheper:
## Summary

Fix the two CWE-78 command-injection findings with a minimal, four-file change.

- Validate `GrepTool.max_results` as a bounded integer before interpolating it into the shell pipeline.
- Stage workspace file content in a temporary directory and transfer it through the terminal's existing `copy_content` API, removing heredoc command construction for all payload sizes.
- Add regression tests proving invalid grep limits never reach the terminal and heredoc-shaped file content is written literally without execution.

## Scope

This PR intentionally excludes the unrelated generic tool-parser recovery and advanced filesystem metadata hardening from #371. The CWE-1188 LocalTerminal finding was already resolved by #375, which removed LocalTerminal.

## Validation

- Black
- isort
- Python 3.12 syntax compilation

### DIFF
--- a/debug_gym/gym/tools/grep.py
+++ b/debug_gym/gym/tools/grep.py
@@ -9,6 +9,8 @@ from debug_gym.gym.tools.toolbox import Toolbox
 @Toolbox.register()
 class GrepTool(EnvironmentTool):
     name: str = "grep"
+    DEFAULT_MAX_RESULTS = 100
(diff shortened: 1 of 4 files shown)


matheper:
Nothing was merged into main here. I merged #376 commit into #371 to preserve both histories while making #371 resulting tree match #376, which caused GitHub to mark #376 as merged

